PR #1, opened 2019-09-20
Phase1

Lokathor:
TODO:
* ?


src/pod.rs
```diff
@@ -0,0 +1,90 @@
...
+unsafe impl Pod for Option<NonZeroU8> {}
+unsafe impl Pod for Option<NonZeroU16> {}
+unsafe impl Pod for Option<NonZeroU32> {}
+unsafe impl Pod for Option<NonZeroU64> {}
+unsafe impl Pod for Option<NonZeroU128> {}
+unsafe impl Pod for Option<NonZeroUsize> {}
+
+unsafe impl<T: 'static> Pod for *mut T {}
+unsafe impl<T: 'static> Pod for *const T {}
+unsafe impl<T: 'static> Pod for Option<NonNull<T>> {}
```
MaulingMonkey: I'm curious as to why the 'static constraints are necessary

Lokathor: I was told to place them in there as added protection against accidentally screwing up a lifetime, or something, so I put it in the `Pod` trait so it ended up being needed on the pointer types too. I'm not super clear on what `+'static` means, I guess just "no borrows"?

I figured that it's generally safer to have more bounds than less, so I didn't really try to argue the point at the time.

Lokathor:
Ah, right, none. I just literally forget that you can do that

src/allocation.rs
```diff
@@ -0,0 +1,134 @@
...
+  if align_of::<A>() != align_of::<B>() {
+    Err((PodCastError::AlignmentMismatch, input))
+  } else if size_of::<A>() != size_of::<B>() {
+    Err((PodCastError::SizeMismatch, input))
+  } else {
+    // Note(Lokathor): This is much simpler than with the Vec casting!
+    let ptr: *mut B = Arc::into_raw(input) as *mut B;
+    Ok(unsafe { Arc::from_raw(ptr) })
+  }
+}
```
MaulingMonkey: This makes me all kinds of nervous.  Specifically if A is stored inside of something like this under the hood:
```rs
// NOTE: repr(rust)
struct ArcRefCounted<A> {
    rc: AtomicUsize,
    data: A,
}
```
The position of rc wouldn't be guaranateed to be compatible between A and B to my understanding.  If this pointer casting *is* sound/guaranteed by the compiler, I'd like to see some references in the comments!

MaulingMonkey: ...the real type is ArcInner and it *is* repr(rust)... paranoia senses tingling...

Lokathor: Yeah, you're right. and `Rc` has this same problem.

MaulingMonkey: This also happens with .as_mut() FWIW but glad I'm not missing anything

src/lib.rs
```diff
@@ -158,11 +158,7 @@ pub fn try_cast_ref<A: Pod, B: Pod>(a: &A) -> Result<&B, PodCastError> {
   if align_of::<B>() > align_of::<A>() && (a as *const A as usize) % align_of::<B>() != 0 {
     Err(PodCastError::TargetAlignmentGreaterAndInputNotAligned)
   } else if size_of::<B>() == size_of::<A>() {
-    Ok(unsafe {
-      (a as *const A as *const B)
-        .as_ref()
-        .unwrap_or_else(|| core::hint::unreachable_unchecked())
-    })
+    Ok(unsafe { &*(a as *const A as *const B) })
```
MaulingMonkey: This also happened with .as_mut() elsewhere in this file FWIW but glad I'm not missing anything

Lokathor:
@MaulingMonkey there was a request for zeroed boxes without using the stack, so i put that in.

src/pod.rs
```diff
@@ -0,0 +1,90 @@
...
+/// big-endian machines do exist (and big-endian is also used for "network
+/// order" bytes).
+///
+/// ## Safety
+///
+/// * The type must be inhabited (eg: no
+///   [Infallible](core::convert::Infallible)).
+/// * The type must allow any bit pattern (eg: no `bool` or `char`).
+/// * The type must not contain any padding bytes (eg: no `(u8, u16)`).
+/// * A struct needs to be `repr(C)`, or a `repr(transparent)` wrapper around a
```
danielhenrymantilla: `#[repr(packed)]` should work too

src/zeroable.rs
```diff
@@ -0,0 +1,117 @@
...
+///   [Infallible](core::convert::Infallible)).
+/// * Your type must be allowed to be an "all zeroes" bit pattern (eg: no
+///   [`NonNull<T>`](core::ptr::NonNull)).
+pub unsafe trait Zeroable: Sized {
+  /// Calls [`zeroed`](core::mem::zeroed).
+  ///
+  /// This is a trait method so that you can write `MyType::zeroed()` in your
+  /// code. It is a contract of this trait that if you implement it on your type
+  /// you **must not** override this method.
+  fn zeroed() -> Self {
```
danielhenrymantilla:
```suggestion
  #[inline]
  fn zeroed() -> Self {
```

src/pod.rs
Lokathor: got it

src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+/// As [`try_cast_slice`], but `mut`.
+#[inline]
+pub fn try_cast_slice_mut<A: Pod, B: Pod>(a: &mut [A]) -> Result<&mut [B], PodCastError> {
+  // Note(Lokathor): everything with `align_of` and `size_of` will optimize away
+  // after monomorphization.
+  if align_of::<B>() > align_of::<A>() && (a.as_ptr() as usize) % align_of::<B>() != 0 {
+    Err(PodCastError::TargetAlignmentGreaterAndInputNotAligned)
+  } else if size_of::<B>() == size_of::<A>() {
+    Ok(unsafe { core::slice::from_raw_parts_mut(a.as_ptr() as *mut B, a.len()) })
+  } else if size_of::<A>() == 0 || size_of::<B>() == 0 {
```
danielhenrymantilla: The case `size_of::<A>() == 0` is acceptable: it can just lead to returning `&mut [] : &mut [B]`, right?

src/allocation.rs
```diff
@@ -0,0 +1,108 @@
...
+    let capacity: usize = input.capacity();
+    // Note(Lokathor): Next we "pre-forget" the old Vec by wrapping with
+    // ManuallyDrop, because if we used `core::mem::forget` after taking the
+    // pointer then that would invalidate our pointer (I think? If not this
+    // still doesn't hurt).
+    let mut manual_drop_vec = ManuallyDrop::new(input);
+    // Note(Lokathor): Finally, we carefully get the pointer we need, cast the
+    // type, and then make a new Vec to return. This works all the way back to
+    // 1.7, if you're on 1.37 or later you can use `Vec::as_mut_ptr` directly.
+    let vec_ptr: *mut A = Vec::as_mut_slice(&mut manual_drop_vec).as_mut_ptr();
```
danielhenrymantilla: Make the deref explicit:
```suggestion
    let vec_ptr: *mut A = Vec::as_mut_slice(&mut *manual_drop_vec).as_mut_ptr();
```

src/lib.rs
Lokathor: I decided to disallow it because this way you can always round-trip a slice or you can't cast it at all, no casts you can't come back from, which feels better somehow.

src/allocation.rs
Lokathor: got it

src/allocation.rs
```diff
@@ -31,6 +33,29 @@ pub fn try_cast_box<A: Pod, B: Pod>(input: Box<A>) -> Result<Box<B>, (PodCastErr
...
 
+/// Allocates a `Box<T>` with all of the contents being zeroed out.
+///
+/// This is not the same as using `Box::new(T::zeroed())`. With this function
+/// the contents of the box never exists on the stack, so this can create very
+/// large boxes without fear of a stack overflow.
+#[inline]
+pub fn try_zeroed_box<T: Zeroable>() -> Result<Box<T>,()>  {
+  let layout = Layout::from_size_align(size_of::<T>(), align_of::<T>()).unwrap();
+  let ptr = unsafe { alloc_zeroed(layout) };
```
MaulingMonkey: "This function is unsafe because undefined behavior can result if the caller does not ensure that layout has non-zero size." ( https://doc.rust-lang.org/beta/std/alloc/trait.GlobalAlloc.html#safety-1 )

I don't see anything guaranteeing layout is !0 size.

Lokathor: frekkin zero sized types

MaulingMonkey: Best bet might be a `#[inline(never)]` helper that's called when `size_of::<T>() == 0` to just construct it on the stack in the more typical manner.

Lokathor: I'll just block it entirely for now. The libstd docs don't actually specify how boxes and zst interact.

danielhenrymantilla:
What about returning `Ok(Box::<T>::new(T::zeroed()))` ?

Lokathor:
That puts T::zeroed on the stack, which is exactly what the goal was to avoid.

You're the second person to ask that one. Do yall read the doc comments?

danielhenrymantilla:
Sorry, I posted a global comment instead of answering to the thread about ZST in `try_zeroed_box`. In that case instead of erroring an `Ok(Box::new(T::zeroed()))` is acceptable since `T` is zero-sized so there is nothing in the stack.

Lokathor:
ahh, hmm, that sounds like a good fix

MaulingMonkey:
```rs
#[inline(never)] fn try_zeroed_box_on_stack<T>() -> Box<T> {
    Ok(Box::new(T::zeroed()))
}
```
Would help ensure it never gets inlined onto the stack for `size_of::<T>() != 0`... although I imagine the branch would get optimized out in release builds, so maybe just a regular fn would be fine?

Lokathor:
If we only use that for zero sized types it doesn't matter if it is "on the stack" since the size is zero anyway

MaulingMonkey:
Right, I'm just being paranoid about the case where non-executing code isn't fully eliminated (debug builds?), which *can* still contribute to stack size as I've found out before (last time I overflowed the stack was thanks to conditional logging macros, that had fixed length char buffers for formatting strings... apparently that's enough to blow the stack even when they're not executing....)

Or to put it a little more concretely, I'd worry:
```rs
if size_of::<T>() != 0 {
    Ok(Box::new(T::zeroed()))
} else {
    ...
}
```
Could still put size_of::<T> bytes on the stack when it's nonzero, despite the branch never executing.

PR #7, opened 2019-12-06
Support casting a slice of ZST’s to an (empty) slice of non-ZST’s

SimonSapin:
That limitation seems unnecessary.

The second commit is more subjective, depending on whether you prefer reducing uses of `unsafe` or reducing uses of `unwrap`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,9 @@ pub use pod::*;
 /// Any ZST becomes an empty slice.
 #[inline]
 pub fn bytes_of<T: Pod>(t: &T) -> &[u8] {
-  try_cast_slice::<T, u8>(core::slice::from_ref(t)).unwrap()
+  unsafe {
+    core::slice::from_raw_parts(t as *const T as *const u8, size_of::<T>())
+  }
 }
 
 /// Re-interprets `&mut T` as `&mut [u8]`.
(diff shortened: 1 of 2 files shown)


Lokathor:
Before I start looking at this too closely, I'll say that the current design is the way it is so that if a slice casts at all you can always round trip it.

thomcc:
Hrm, is that legal? If the ZST is underaligned compared to the destination type, `result_slice.as_ptr()` wouldn't be aligned.

That said, I don't know if that's a thing that actually matters. (I guess maybe this is somewhat equivalent to: is `slice::from_raw_parts(ptr, 0)` safe even if `ptr` is garbage, which isn't something I know the answer to)

Lokathor:
Miri thinks it is not legal, https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bd561d8d2452317af51c2a84eb477eac

SimonSapin:
> so that if a slice casts at all you can always round trip it.

Oh ok, I didn’t realize that.

> Hrm, is that legal?

By “that” do you mean this PR or the round-trip constraint? This PR is not changing the existing alignment checks.

Lokathor:
Unfortunately, I think I'm not going to take in this PR at this time.

I'd be willing to reconsider the issue at a later time if there was wider demand that this was an important ability, but otherwise I'd like for the casting to always be either a lossless-round-trip on success or a failure. Also for ZST to generally stay a little apart from the sized type world.

SimonSapin:
That’s fine, feel free to close. I submitted this as a drive-by change that seemed like an improvement, but I don’t particularly need it.

PR #8, opened 2019-12-06
Add from_bytes, from_bytes_mut, try_from_bytes, try_from_bytes_mut

SimonSapin:
This can be useful when receiving a bytes buffer from I/O

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,6 +81,62 @@ pub fn bytes_of_mut<T: Pod>(t: &mut T) -> &mut [u8] {
   try_cast_slice_mut::<T, u8>(core::slice::from_mut(t)).unwrap_or(&mut [])
 }
 
+/// Re-interprets `&[u8]` as `&T`.
+///
+/// ## Panics
+///
+/// This is [`try_from_bytes`] with an unwrap.
+#[inline]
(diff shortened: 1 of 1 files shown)


SimonSapin:
Open question: should these functions succeed if more bytes than necessary are passed? As proposed in https://internals.rust-lang.org/t/pre-rfc-v2-safe-transmute/11431

Lokathor:
Well that answer is easy: Fail early on unexpected input.

If the user wants to cast a slice it's up to them to make sure that the slice has the correct number of bytes. Rust has several ways to cut slices down to the desired size, so it's up to you to call `.chunks().next()` or `[..size_of<Target>()]` or `split_at(size_of::<Target>())` or however else.

This might sound like idle theory crafting, but when writing my PNG parser and doing a little too much copy paste I had some `&[u8]` values that I was casting to `RGBA8` in one method and then in the copy-paste version I was going to `RGB8` (no A). I was using a variant of these at the time (`cast_slice(bytes)[0]`) and the fact that I was handing over 4-byte slices and wanting back 3-byte values caught what could have been a very tricky to hunt down bug.

SimonSapin:
I tend to agree, and it’s what this PR does. I mentioned the alternative since at least some people seem to prefer that.

Lokathor:
v1.1.0 is published!

PR #15, opened 2020-01-10
Add TransparentWrapper marker trait

thomcc:
That transmute_copy bothers me, and I'd like have a "do sizes match" sanity check for broken impls... But supporting &[T] to &SomeStruct([T]) is too useful to give up IMO.

Unsure on where wrap_ref/wrap_mut should go. I don't really like "hey don't implement this" functions on the trait, but maybe they'd be better there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,6 +66,9 @@ pub use pod::*;
 mod contiguous;
 pub use contiguous::*;
 
+mod transparent;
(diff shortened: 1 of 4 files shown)


thomcc:
The last commit here arguably should be another PR, but I think I'm the one who wrote the doctests anyway, so whatever.

PR #24, opened 2020-07-24
add additional license options

Lokathor:
Hello all, we're adding the standard `Apache-2.0 OR MIT` license options to the crate, in addition to the existing `Zlib` license option. I expect this to be agreeable with everyone since Rust uses these licenses all the time, but I still do need past code contributors to sign off on this:
* [x] @thomcc 
* [x] @SimonSapin 
* [x] @HeroicKatora 

so just comment something like "i agree", or whatever.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,14 +1,14 @@
 [package]
 name = "bytemuck"
 description = "A crate for mucking around with piles of bytes."
-version = "1.3.1-alpha.0"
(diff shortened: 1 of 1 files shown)


thomcc:
Fine by me.

HeroicKatora:
I agree to license my contributions under Apache-2.0 OR MIT as well.

SimonSapin:
I agree

PR #30, opened 2020-08-17
add basic derive macro for Pod, Zeroable and TransparentWrapper for structs

### LINKED ISSUE #4: Derives for traits?
Basically I'm wondering if it would be useful to have derive macros for `Pod` and `Zeroable`.  They could be hidden behind a feature flag so as to not make things slower in the simple case (similar to `serde_derive`), but could also automatically check that the various constraints for these traits are upheld.

icewind1991:
Adds derive macros for `Zeroable` and `Pod` for structs.

The macro is exported behind the `derive` feature flag (same way as serde does)

- The `Zeroable` macro checks that all fields are `Zeroable`.
- The `Pod` macro checks that all fields are `Pod`, there is no padding bytes and that the struct is `#[repr(C)]` or `#[repr(packed)]`

This currently only supports deriving for structs, deriving `Zeroable` would be possible in the future.

Fixes #4

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -36,12 +36,16 @@ pub fn derive_transparent(
 }
 
 fn derive_marker_trait<Trait: Derivable>(input: DeriveInput) -> TokenStream {
-  derive_marker_trait_inner::<Trait>(input).unwrap_or_else(|err| err)
+  derive_marker_trait_inner::<Trait>(input).unwrap_or_else(|err| {
+    quote! {
+      compile_error!(#err);
+    }
+  })
 }
 
 fn derive_marker_trait_inner<Trait: Derivable>(
   input: DeriveInput,
-) -> Result<TokenStream, TokenStream> {
+) -> Result<TokenStream, &'static str> {
   let name = &input.ident;
 
   let (impl_generics, ty_generics, where_clause) =
@@ -51,9 +55,7 @@ fn derive_marker_trait_inner<Trait: Derivable>(
   let fields = if let Data::Struct(DataStruct { fields, .. }) = &input.data {
     fields
   } else {
-    return Err(quote! {
-      compile_error!("deriving this trait is only supported for structs");
-    });
+    return Err("deriving this trait is only supported for structs");
   };
 
   let trait_ = Trait::ident();
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -1,11 +1,37 @@
+//! Derive macros for [bytemuck](https://docs.rs/bytemuck) traits
+
 mod traits;
 
 use proc_macro2::TokenStream;
 use quote::quote;
-use syn::{parse_macro_input, spanned::Spanned, Data, DataStruct, DeriveInput};
+use syn::{parse_macro_input, DeriveInput};
 
 use crate::traits::{Derivable, Pod, TransparentWrapper, Zeroable};
 
+/// Derive the `Pod` trait for a struct
+///
+/// The macro ensures that the struct follows all the the safety requirements
+/// for the `Pod` trait.
+///
+/// The following constraints need to be satisfied for the macro to succeed
+///
+/// - All fields in the struct must implement `Pod`
+/// - The struct must be `#[repr(C)]` or `$[repr(transparent)]`
+/// - The struct must not contain any padding bytes
+/// - The struct contains no generic parameters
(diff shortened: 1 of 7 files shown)


Shnatsel:
Now that `rgb` crate requires these trait bounds, it'd be nice to have a safe way to derive them.

@danielhenrymantilla do I recall correctly that you've come up with a way to derive these via macros-by-example only?

Lokathor:
I'm very interested in the idea of this PR, but I won't have the time to look until later today, and I'm also a little nervous about putting proc-macros near unsafe code since I don't really use them much, so I'm going to ask around for extra eyes to look at this to make sure it's all good.

icewind1991:
added a macro for deriving `TransparentWrapper`

it checks that the struct is `#[repr(transparent)]` and contains the a field with the type specified in the trait bounds.

If the struct only contains a single field it will automatically use the type of that field for the train bound, if there is more than one field (extra zero-sized fields) you'll need to specify the type using `#[transparent(T)]`.

For example:

```rust
#[derive(TransparentWrapper)]
#[repr(transparent)]
struct TransparentSingle {
  a: u16,
}

#[derive(TransparentWrapper)]
#[repr(transparent)]
#[transparent(u16)]
struct TransparentWithZeroSized {
  a: u16,
  b: (),
}
```

note that this macro is more strict that the requirements of the trait, since the trait allows for "nested" transparent wrappers:

```rust
#[repr(transparent)]
struct A(u16);

#[repr(transparent)]
struct B(A);

unsafe impl TransparentWrapper<u16> for B{};
```

the macro currently rejects this since I can't think of a good way to verify the constraints while allowing this

Lokathor:
it looks like some of the output code is calling std::mem::size_of ? That at least can be core::mem::size_of.

Shnatsel:
https://crates.io/crates/zerocopy-derive solves largely the same problem, perhaps some techniques from it can be adopted here.

icewind1991:
> it looks like some of the output code is calling std::mem::size_of ? That at least can be core::mem::size_of.

Fixed

icewind1991:
running 

[cargo expand](https://github.com/dtolnay/cargo-expand/) --test basic

from the derive folder is probably the easiest way to check what checks are being generated by the macro

derive/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+
+/// Derive the `Pod` trait for a struct
+///
+/// The macro ensures that the struct follows all the the safety requirements
+/// for the `Pod` trait.
+///
+/// The following constraints need to be satisfied for the macro to succeed
+///
+/// - All fields in the struct must implement `Pod`
+/// - The struct must be `#[repr(C)]` or `$[repr(transparent)]`
```
LPGhatguy, line 19:
```suggestion
/// - The struct must be `#[repr(C)]` or `#[repr(transparent)]`
```

Typo?

derive/src/traits.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    Self::get_wrapper_type(&input.attrs, fields).map(|ty| quote!(<#ty>))
+      .ok_or("when deriving TransparentWrapper for a struct with more than one field you need to specify the transparent field using #[transparent(T)]")
+  }
+
+  fn struct_asserts(input: &DeriveInput) -> Result<TokenStream, &'static str> {
+    let fields = get_struct_fields(input)?;
+    let wrapped_type = match Self::get_wrapper_type(&input.attrs, fields) {
+      Some(wrapped_type) => wrapped_type.to_string(),
+      None => return Err(""), /* other code will already reject this derive */
```
LPGhatguy, line 101: Should this be an `unreachable!()` instead, to better signal the invariant?

derive/src/traits.rs
```diff
@@ -0,0 +1,209 @@
...
+      _ => {
+        Err("TransparentWrapper requires the struct to be #[repr(transparent)]")
+      }
+    }
+  }
+}
+
+fn get_struct_fields<'a>(
+  input: &'a DeriveInput,
+) -> Result<&'a Fields, &'static str> {
```
LPGhatguy, line 130: Does the lifetime `'a` need to be named in this function?

Lokathor: since the &'static str has a named lifetime it's good to name the other lifetime, even if not absolutely required.

PR #35, opened 2020-08-24
(WIP) Improve error messages from derive macros

### LINKED ISSUE #4: Derives for traits?

LPGhatguy:
Fixes to help close #4.

Error messages from padding for deriving `Pod` before looked like this:

before

...and now they look like this, which is a small improvement:

after

I couldn't figure out a way to make the second type have a nicer name, but having any name at all can hint at what's going wrong I guess.

### DIFF
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -210,17 +210,16 @@ fn generate_assert_no_padding(
   input: &DeriveInput,
 ) -> Result<TokenStream, &'static str> {
   let struct_type = &input.ident;
-  let span = input.span();
+  let span = input.ident.span();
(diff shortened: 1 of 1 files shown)


Lokathor:
i love it.

Lokathor:
I'll publish another alpha in a bit.

I'll do the full release of the derives as 1.0 on Thursday if no other problems happen before then.

PR #38, opened 2020-09-06
Add more tests, overhaul CI, and fix plug soundness hole in `bytemuck::offset_of!`

thomcc:
I gave you a heads up in discord about this. Basically, I had a bunch of 80% finished work in `git stash`es, since I keep meaning to give bytemuck a bit more attention. Anyway I found some time and cleaned them up.

... Which is to say, this PR is a bit of a hodge-podgey mess of things that are useful but not meaningfully related. It's not *too* big, but it's also not ideal to review.

That said, what's here is pretty nice maybe? In order of smallest to largest:

1. (first commit) Some minor doc fixing I noticed a while ago where you call types `T` and `U` but the function takes `A` and `B` type params. IMO in an ideal world these would be `From` and `To` but at a minimum the docs should use the right names.

2. (third commit) Add some more unit tests. I ran a coverage scan of `bytemuck` at one point and it only has like 50%. Many of it's functions are very small and unlikely to hold bugs, but it doesn't really feel acceptable to have a crate with so much unsafe to have low coverage. Eventually I'll probably add more.

3. (fourth commit) Totally rewrite the CI. I was concervative in what I forced on you here, but now it supports:
    - All the same stuff it was doing, e.g. test on 1.34.0, stable, beta, nightly on linux-x86_64.
    - Direct (non-`cross`) tests on `macOS-x86_64`, `windows-{x86_64,i686}-{msvc,gnu}`.
    - `cross`-based tests on `linux-i686`, and 32-bit/64-bit mips (mips is obscure but is a big endian system which doesn't cause much trouble when run in `cross` *cough unlike ppc cough*)
    - Running tests under `miri`, `asan`, `msan`, and `lsan`.

    - I didn't add anything that might be a little controversial, like lint/fmt/coverage/doc-check, but you can take a look at what I considered here: https://gist.github.com/thomcc/4d37f0ae675b3a2cba9ca5a7e2ba8baa. Of these IMO doc check is the most useful, since it makes it easier to be confident about cross-referencing in the docs.

	- (Oh and ofc I totally eyeballed this and can't test it locally so I'll probably have to add some followup commits to ensure it compiles. As you do).

4. (third commit) Fix the soundness hole in `bytemuck::offset_of!`, and overhaul it's docs. This is done using `#[forbid(safe_packed_borrows)]` inside the macro body. This works at least back to the MSRV, didn't check any further.

    - It comes with `compile_fail` tests (to ensure soundness), and an explanation that basically boils down to "No, this doesn't just mean you should wrap the macro in `unsafe`".

    - This is p. straightforward, but probably harder to review than the others since it's an API change that impacts safety.

### DIFF
--- a/src/offset_of.rs
+++ b/src/offset_of.rs
@@ -17,16 +17,6 @@
 /// type used is `repr(Rust)` then they're *not* fixed across compilations or
 /// compilers.
 ///
-/// **CAUTION:** It is **unsound** to use this macro with a `repr(packed)` type.
-/// Currently this will give a warning, and in the future it will become a hard
-/// error.
-/// * See [rust-lang/rust#27060](https://github.com/rust-lang/rust/issues/27060)
-///   for more info and for status updates.
-/// * Once this issue is resolved, a future version of this crate will use
-///   `raw_ref` to correct the issue. For the duration of the `1.x` version of
-///   this crate it will be an on-by-default cargo feature (to maintain minimum
-///   rust version support).
-///
 /// ## Examples
 ///
 /// ### 3-arg Usage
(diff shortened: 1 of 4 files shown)


src/offset_of.rs
```diff
@@ -64,20 +54,77 @@
...
+/// # Usage with `#[repr(packed)]` structs
+///
+/// Attempting to compute the offset of a `#[repr(packed)]` struct with
+/// `bytemuck::offset_of!` requires an `unsafe` block. We hope to relax this in
+/// the future, but currently it is required to work around a soundness hole in
+/// Rust (See [rust-lang/rust#27060]).
+///
+/// [rust-lang/rust#27060]: https://github.com/rust-lang/rust/issues/27060
+///
+/// <p style="background:rgba(255,181,77,0.16);padding:0.75em;">
```
thomcc, line 67: Note: This is kind of gross looking in the docs, but looks very nice when rendered: https://imgur.com/a/blwzDfE

Lokathor (approved):

Lokathor:
So this is just a patch release it seems like, no new API.

PR #43, opened 2020-10-12
Prevent `try_zeroed_box<T>()` from reserving `size_of<T>()` space on the stack.

Kimundi:


Lokathor:
@thomcc does that miri error seem normal? Should we change the CI to use a `cargo clean` first like it says?

PR #45, opened 2020-10-12
Emit padding-asserting code that doesn't trigger clippy::identity_op

### LINKED ISSUE #44: derive(Pod) generates code that triggers warn-by-default clippy::identity_op
Hi! Thanks for bytemuck :heart:  It is a pleasure to work with (compared to my previous unsafe versions of `bytes_of` and `from_bytes` ).

When switching to it, I managed to trigger https://rust-lang.github.io/rust-clippy/master/index.html#identity_op in the code generated by the derives.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=%23%5Brepr(C)%5D%0A%23%5Bderive(Debug%2C%20Clone%2C%20Copy%2C%20PartialEq%2C%20Default%2C%20bytemuck%3A%3AZeroable%2C%20bytemuck%3A%3APod)%5D%0Apub%20struct%20Vec2%20%7B%0A%20%20%20%20x%3A%20f32%2C%0A%20%20%20%20y%3A%20f32%2C%0A%7D%0A%0Afn%20main()%20%7B%7D

(run `clippy` on tthe play pen)

The culprit seems to be this generated code, emitted as a static check before the unsafe impl is emitted:

```rust
const _: fn() =
    ||
        {
            struct TypeWithoutPadding([u8; 0 + ::core::mem::size_of::<f32>() +
                                               ::core::mem::size_of::<f32>()]);
            let _ = ::core::mem::transmute::<Vec2, TypeWithoutPadding>;
        };
```
Seems to be trivially fixable by checking the number of fields beforehand in https://github.com/Lokathor/bytemuck/blob/main/derive/src/traits.rs#L209

I can submit a fix, if you desired.

yanchith:
Not sure this was worth it, but here goes.. 😹 

Fixes #44

### DIFF
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -213,9 +213,16 @@ fn generate_assert_no_padding(
   let span = input.ident.span();
   let fields = get_struct_fields(input)?;
 
-  let field_types = get_field_types(&fields);
-  let size_sum =
-    quote_spanned!(span => 0 #( + ::core::mem::size_of::<#field_types>() )*);
+  let mut field_types = get_field_types(&fields);
+  let size_sum = if let Some(first) = field_types.next() {
+    let size_first = quote_spanned!(span => ::core::mem::size_of::<#first>());
+    let size_rest =
+      quote_spanned!(span => #( + ::core::mem::size_of::<#field_types>() )*);
+
+    quote_spanned!(span => #size_first#size_rest)
+  } else {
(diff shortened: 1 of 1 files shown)


Lokathor:
Right, this CI error is the same as the miri error in the other PR. It's probably something we need to fix in the CI.

Otherwise this looks good!

Issue #49, opened 2021-01-10
Introduce a function to align a slice to the right alignment


Kerollmops:
Hello @Lokathor,

I wanted [to switch from zerocopy to bytemuck for multiple reasons](https://github.com/Kerollmops/heed/issues/82), and one of the missing features for both crates is the ability to realign a slice of type `A` into a `Vec` of type `B`, which is allowed to allocate. What do you think about introducing this to this crate (under the `allocation` feature) and introducing that for `Box`?

This function would be preferable not to restrict the size (where the length of the input must be the same as the output `Vec`). It would make this function useless for when you want to transpose differently sized types (e.g. `u8` into `u16`) like for [the `try_cast_vec` function](https://docs.rs/bytemuck/1.4.1/bytemuck/allocation/fn.try_cast_vec.html).

Another solution would be to make [the `try_cast_vec` function](https://docs.rs/bytemuck/1.4.1/bytemuck/allocation/fn.try_cast_vec.html) change the length and capacity of the `Vec`. This seems a lot better but does seem to be a breaking change in the sense that more results could return an `Ok(Vec)` instead of an error and that the `Vec` is no more of the same length than the input one.

```rust
// An idea of the possible function signature
fn try_cast_slice_realign_vec<A: Pod, B: Pod>(input: &[A]) -> Result<Vec<B>, PodCastError>;

// Another even better function signature which will allocate only if alignment is invalid
// Note that Cow is in the alloc Rust library
fn try_cast_slice_realign_cow<A: Pod, B: Pod>(input: &[A]) -> Result<Cow<[B]>, PodCastError>;
```

Could maybe be implemented this way:

```rust
let bytes = bytes_of(input);
let len = input.len();
let elem_size = mem::size_of::<B>();

// ensure that it is the alignment that is wrong
// and the length is valid
if len % elem_size == 0 && align_of::<A>() != align_of::<B>() {
    let elems = len / elem_size;
    let mut vec = Vec::<B>::with_capacity(elems);

    unsafe {
        let dst = vec.as_mut_ptr() as *mut u8;
        ptr::copy_nonoverlapping(bytes.as_ptr(), dst, len);
        vec.set_len(elems);
    }

    Ok(Cow::Owned(vec))
}
```

Lokathor:
The capacity and alignment of the vec's allocation can't change during casting, it causes UB in the allocator.

[...]

PR #50, opened 2021-01-13
pod_collect_to_vec

Lokathor:
Closes https://github.com/Lokathor/bytemuck/issues/49

@Kerollmops is this what you had in mind?


Lokathor:
Note: the failures are in miri but this adds no new unsafe code, so we'll ignore the miri failures.

Kerollmops:
This is indeed what I had in mind, I was more like: if the size is invalid we return an error and only if the alignment is wrong we allocate (using a Cow to express that). But this is enough for me to do that myself with this function.

Thank you very much!

Lokathor:
released in `1.5.0`

Issue #34, opened 2020-08-23
Use trybuild crate on the derive macros


Lokathor:
https://github.com/dtolnay/trybuild

This would let us declare things that _should fail_ and then check that they really do fail.

Specifically, we could make a test such as

```rust
#[derive(Zeroable, Pod)]
pub struct Bad(u16, u8);
```
And that should fail to derive, of course.

mobile-bungalow:
I Implemented these tests over on my [branch](https://github.com/mobile-bungalow/bytemuck/tree/trybuild_tests) with one major hiccup. Trybuild checks the build output against the previous stderr to make sure everything failed for the same reasons. This means github actions will permanently fail on nightly builds as the output is different. [There is an issue for this](https://github.com/dtolnay/trybuild/issues/79). 

If you are to move forward with this: would you prefer to suppress checking the previous output or to copy in a separate set of stderr files in CI? Are there any other solutions?

Lokathor:
It appears that, through a fluke of timing, they closed the trybuild issue as a "won't fix" after you posted this but before i saw your comment.

Let's just suppress the previous output check. The output doesn't really matter as long as the derive fails with *some* sort of error.

PR #54, opened 2021-02-08
Added trybuild tests

mobile-bungalow:
## Changes

1.) removed old derive test. converting it into a trybuild test.
2.) added four new trybuild tests, one for each derive macro covering the cases when it should fail.
3.) updated CI to run the tests on all toolchains except MIRI which cannot run trybuild tests. The tests ignore any .stderr output that was left in the testing subdirectory and only fail if the compilation did not exit as expected.
4.) added *.stderr to the gitignore to prevent people from polluting the CI environment with build artifacts.

## Notes
I attempted to modify CI to select and move the proper stderr files into place before trybuild testing. It worked but seemed to be a recipe for creating a flakey build every 6 weeks as well as being hard to maintain. I've opted for the method we discussed in #34 .  

currently the memory san test fails. I could use help finding out why. this is likely a blocker.


tests/trybuild_tests_src/contiguous_derive_failure.rs
```diff
@@ -0,0 +1,17 @@
+use bytemuck::Contiguous;
```
mobile-bungalow, line 1: Just now realizing that these tests won't fail unless the derive macro fails _all_ cases. It might be worth it to only run these tests in CI on stable and keep the stderr files up to date.

Lokathor:
Hey @mobile-bungalow I'm unclear of the PR status here. Do you want to continue this PR?

mobile-bungalow:
I Think this PR is pretty dead in the water. I encountered too many issues that made the tests flakey. Closing this.

PR #58, opened 2021-03-21
[Feature] extend `TransparentWrapper` conversion functions

LU15W1R7H:
This PR extends the `TransparentWrapper` trait by adding new conversion functions.
Namely:

These three missing `wrap_{}` functions.
- `TransparentWrapper::wrap(s: Inner) -> Self`
- `TransparentWrapper::wrap_slice(s: &[Inner]) -> &[Self]`
- `TransparentWrapper::wrap_slice_mut(s: &mut [Inner]) -> &mut [Self]`

These five `unwrap_{}` functions, which are the counterparts of the `wrap_{}` functions. They just invert the conversion.
- `TransparentWrapper::unwrap(self) -> Inner`
- `TransparentWrapper::unwrap_ref(&self) -> &Inner`
- `TransparentWrapper::unwrap_mut(&mut self) -> &mut Inner`
- `TransparentWrapper::unwrap_slice(s: &[Self]) -> &[Inner]`
- `TransparentWrapper::unwrap_slice_mut(s: &mut [Self]) -> &mut [Inner]`

This is useful for implementing `Zeroable` and `Pod` on types from extern crates in order to use `bytemuck::{}` functions like `bytemuck::cast_slice` using the new-ype pattern and `TransparentWrapper`.

The goal of this PR is to support all `bytemuck::{}` functions.

This is possible with this PR:
```rust
  // An external type defined in a different crate.
  #[derive(Copy, Clone, Default)]
  struct Foreign(u8);

  use bytemuck::TransparentWrapper;

  #[derive(Copy, Clone)]
  #[repr(transparent)]
  struct Wrapper(Foreign);

  unsafe impl TransparentWrapper<Foreign> for Wrapper {}

  // Traits can be implemented on crate-local wrapper.
  unsafe impl bytemuck::Zeroable for Wrapper {}
  unsafe impl bytemuck::Pod for Wrapper {}

  let _: u8 = bytemuck::cast(Wrapper::wrap(Foreign::default()));
  let _: Foreign = Wrapper::unwrap(bytemuck::cast(u8::default()));

  let _: &u8 = bytemuck::cast_ref(Wrapper::wrap_ref(&Foreign::default()));
  let _: &Foreign = Wrapper::unwrap_ref(bytemuck::cast_ref(&u8::default()));

  let _: &mut u8 = bytemuck::cast_mut(Wrapper::wrap_mut(&mut Foreign::default()));
  let _: &mut Foreign = Wrapper::unwrap_mut(bytemuck::cast_mut(&mut u8::default()));

  let _: &[u8] = bytemuck::cast_slice(Wrapper::wrap_slice(&[Foreign::default()]));
  let _: &[Foreign] = Wrapper::unwrap_slice(bytemuck::cast_slice(&[u8::default()]));

  let _: &mut [u8] = bytemuck::cast_slice_mut(Wrapper::wrap_slice_mut(&mut [Foreign::default()]));
  let _: &mut [Foreign] = Wrapper::unwrap_slice_mut(bytemuck::cast_slice_mut(&mut [u8::default()]));

  let _: &[u8] = bytemuck::bytes_of(Wrapper::wrap_ref(&Foreign::default()));
  let _: &Foreign = Wrapper::unwrap_ref(bytemuck::from_bytes(&[u8::default()]));

  let _: &mut [u8] = bytemuck::bytes_of_mut(Wrapper::wrap_mut(&mut Foreign::default()));
  let _: &mut Foreign = Wrapper::unwrap_mut(bytemuck::from_bytes_mut(&mut [u8::default()]));

  // not sure if this is the right usage
  let _ = bytemuck::pod_align_to::<_, u8>(Wrapper::wrap_slice(&[Foreign::default()]));
  // counterpart?

  // not sure if this is the right usage
  let _ = bytemuck::pod_align_to_mut::<_, u8>(Wrapper::wrap_slice_mut(&mut [Foreign::default()]));
  // counterpart?
```

This PR also includes a test, which uses the exact example code from above. This is for MIRI to check for any potential UB.
As you can see, there currently isn't any checking done on the output of the functions. This is something left todo.

Can somebody please check the soundness of the implementation, because I have very little expertise with unsafe code.

I've written up some docs on how this trait can be used in combination with the newtype pattern to use the bytemuck functions on foreign types from external crates.

I also adjusted `rustfmt.toml` in the first commit, because of the deprecation of `merge_imports = true`. It is being replaced by `imports_granularity = "Crate"`. Furthermore I ran `cargo +nightly fmt` to format the codebase. I hope these two changes are fine.

I'm open to suggestions!

This is my first PR ever 🥳 .

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -12,5 +12,5 @@ use_small_heuristics = "Max"
 
 # Unstable
 format_code_in_doc_comments = true
-merge_imports = true
+imports_granularity = "Crate"
 wrap_comments = true
--- a/src/contiguous.rs
+++ b/src/contiguous.rs
@@ -85,8 +85,8 @@ pub unsafe trait Contiguous: Copy + 'static {
   /// `#[repr(Int)]` or `#[repr(C)]` attribute, (if it does not, it is
   /// *unsound* to implement `Contiguous`!).
   ///
-  /// - For `#[repr(Int)]`, use the listed `Int`. e.g. `#[repr(u8)]` should
-  ///   use `type Int = u8`.
+  /// - For `#[repr(Int)]`, use the listed `Int`. e.g. `#[repr(u8)]` should use
+  ///   `type Int = u8`.
   ///
   /// - For `#[repr(C)]`, use whichever type the C compiler will use to
   ///   represent the given enum. This is usually `c_int` (from `std::os::raw`
--- a/src/offset_of.rs
+++ b/src/offset_of.rs
@@ -25,7 +25,9 @@
 /// # use bytemuck::offset_of;
 /// // enums can't derive default, and for this example we don't pick one
 /// enum MyExampleEnum {
-///   A, B, C,
+///   A,
+///   B,
+///   C,
 /// }
 ///
 /// // so now our struct here doesn't have Default
@@ -65,11 +67,11 @@
 /// [rust-lang/rust#27060]: https://github.com/rust-lang/rust/issues/27060
 ///
 /// <p style="background:rgba(255,181,77,0.16);padding:0.75em;">
-/// <strong>Warning:</strong> This is only true for versions of bytemuck > 1.4.0.
-/// Previous versions of
+/// <strong>Warning:</strong> This is only true for versions of bytemuck >
+/// 1.4.0. Previous versions of
 /// <code style="background:rgba(41,24,0,0.1);">bytemuck::offset_of!</code>
-/// will only emit a warning when used on the field of a packed struct in safe code,
-/// which can lead to unsoundness.
+/// will only emit a warning when used on the field of a packed struct in safe
+/// code, which can lead to unsoundness.
 /// </p>
 ///
 /// For example, the following will fail to compile:
@@ -91,7 +93,8 @@
 /// ```compile_fail
 /// # #[repr(C, packed)] #[derive(Default)] struct Example { field: u32 }
 /// // Still doesn't compile:
-/// #[allow(safe_packed_borrows)] {
+/// #[allow(safe_packed_borrows)]
+/// {
 ///   let _offset = bytemuck::offset_of!(Example, field);
 /// }
 /// ```
--- a/src/transparent.rs
+++ b/src/transparent.rs
@@ -6,7 +6,8 @@ use super::*;
 /// This allows safely transmuting between the `Inner` type and the
 /// `TransparentWrapper` type.  
 /// Functions like `wrap_{}` convert from the inner
-/// type to the wrapper type.
+/// type to the wrapper type and `unwrap_{}` functions do the inverse conversion
+/// from the wrapper type to the inner type.
 ///
 /// # Safety
 ///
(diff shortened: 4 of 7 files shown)


Lokathor:
This looks correct.

I'll have @thomcc take a look too, because they wrote this trait.

thomcc (commented):
Up on line 28 you should add this method to the list of methods that may not be overwridden.

Otherwise it seems fine.

LU15W1R7H:
I've now also added the mutable counterpart (`wrap_slice_mut`) to `wrap_slice`.

Instead of mentioning all 4 methods in the safety contract doc, I changed it to say "All functions on `TransparentWrapper` may not be overwritten."

LU15W1R7H:
I've decided to extend this PR some more, therefore I converted it into a draft.
I would like to add all conversions needed, to support all bytemuck functions.

LU15W1R7H:
I've now rewritten and extended the PR quite a bit. Please take a look at it again :).

I would also like to introduce the `wrap` and `unwrap` conversions functions.
This would be nice, because than the conversion for `bytemuck::cast` would look the same as all other conversions.
We could probably just do a transmute.
This would require a `Sized` bound on the function because we need to return a statically sized type.

Is this desirable?

Lokathor:
I believe that the transmute_copy trick used to wrap can also be used to unwrap.

LU15W1R7H:
Okay, another rather big overhaul of this PR is done 😄 .

I adjusted the docs a little to be more concise. Also I renamed `Wrapped` to `Inner` to make it visually more distinct from `Wrapper`.

I've also added the `wrap` and `unwrap` functions. I didn't use the `transmute_copy` **trick** because I don't have to transmute any pointers since there is a `Sized` bound on `Self` and `Inner`. This is needed because rust needs `Sized` functions arguments and return types.

I also added a test which currently just checks for UB using MIRI. Thanks to this I already catched a bug causing UB.
There isn't any actual checking done on the results of the functions.

thomcc:
I'm not sure `unwrap` is the right term for these, given that that terminology is already used very commonly in Rust to mean something different...

LU15W1R7H:
@thomcc I agree, but I haven't come up with a better fitting terminology. It is a wrapper after all :D.
I'm open for suggestions.

LU15W1R7H:
I've now added some docs and a example mentioning how `TransparentWrapper` can be used to implement traits like `Zeroable` and `Pod` on external types.

LU15W1R7H:
This PR is no longer a draft! I'm pretty happy with it.

Now I'm waiting for a review and some feedback on what's left to do :).

Lokathor:
I feel like the derive should continue to fail even then. You can still write it manually. The fact that you didn't just derive it is a good time to say "i know the base type doesn't impl the trait but i trust i can bend the rule in this case"

LU15W1R7H:
@Lokathor Hmm, It's just unfortunate, that in this case you can't have the derive check, if the safety guarantees hold.

Does the current proc-macro implementation check all aspects of the unsafe contract? It should, because otherwise there would be a possibility of introducing UB without having an unsafe block...

If the derive can check all aspects of the unsafe contract, then you shouldn't need to think about bending the rules, because it's guaranteed to be safe and sound :).
I'm just not sure if these new restrictions are easily checkable.

Lokathor:
It may be possible to fix. We could probably finish off the PR noe and file an issue for the derives later.

LU15W1R7H:
Sounds good to me!

src/lib.rs
```diff
@@ -30,6 +30,17 @@
 //! these traits are `unsafe`, and you should carefully read the requirements
 //! before adding the them to your own types.
 //!
+//! ## Using Types from Extern Crates
```
Lokathor, line 33: I don't think that we should document this idea.

It's possible, but it's kinda extra dodgy, particularly because it runs explicitly against the claimed rules of Pod and Zeroable.

LU15W1R7H: Yeah, I get your point. It's getting kinda convoluted.

But I don't know how obvious this use of `TransparentWrapper` is. It is a pretty nice feature to have, so people should know about it.

Maybe we need to reformulate the docs some more? I'll give it a try and if it looks bad, we probably have to remove it...

What is most desirable?
- Mentioning `TransparentWrapper` everywhere where it's relevant (so also in `Pod`).
- Being subtle about it and just mentioning it in it's own docs.
- Not mention it at all.

thomcc: I don't think we should mention it, since it's not sound unless you pin the version exactly. If other crates haven't opted into the bytemuck requirements, they're free to break them at any time without a semver-major change.

Lokathor: Yeah, the reason that the docs for Pod and Zeroable require that the field type also is Pod or Zeroable is because then SemVer kicks in, specifically that it's a SemVer Break to remove a trait impl.

If you newtype a struct and declare the wrapper type Pod, then the inner type adds a `&'static str` field in a minor update (not unreasonable!) then suddenly you're hosed.

So I think that this should stay as a "secret thing the type system technically allows but officially you shouldn't so legally you can't complain when this blows up in your face (and it easily might)" sort of thing.

thomcc:
I think I'm not really a fan of having this many functions on the trait. It would probably be better to move some/most of them to free functions, since at this point there's a real risk of collision with inherent or deref impls, especially with the generic naming scheme...

src/lib.rs
LU15W1R7H: Okay, I see. That's a reasonable argument.
In this case I'm going to remove all mentions of this.

LU15W1R7H:
@thomcc
I'm not quite sure how collisions might happen?
Is it enough to change the `self` to `Self` in the arguments of `unwrap`, `unwrap_ref` and `unwrap_mut`?

LU15W1R7H:
Anything left to do?

Lokathor:
I haven't had time to get to this. My "weekend" is Sun/Mon, so I'll try to get this merged.

Lokathor (approved):
I'm going to approve and merge now, and I'll think hard for a bit about if there's a better name than `unwrap` to use.

I think I also want to sort out how we're going to adapt to min_const_generics being available before the next crates.io release.

LU15W1R7H:
Yeah, my first PR was merged 🥳. Thanks :)

PR #59, opened 2021-03-26
Implement Pod and Zeroable for arrays of any size

zakarumych:
Starting with Rust 1.51 it is possible to use const generics in limited way.
But it's sufficient to implement traits for arrays of all sizes.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -11,7 +11,7 @@ fn main() {
   };
 
   if minor >= 51 {
-    println!("cargo:rustc-cfg=const_generics_mvp");
+    println!("cargo:rustc-cfg=min_const_generics");
   }
 }
 
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,45 @@
+use std::{
+  env,
+  process::Command,
+  str::{self, FromStr},
+};
+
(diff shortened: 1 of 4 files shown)


Lokathor:
I agree that we should do this, but I think it should be via explicit opt-in (a feature such as `rustc_1_51`), not automatic detection.

zakarumych:
Automatic detection would allow users to not bother about it in the future.

For instance, I wouldn't like to go and activate 'rust_1_26' in older crates to make them implement some traits for u128, especially if those are not my direct dependencies. I like to simply assume they all support it already.

This is how serde handles new API and it works great.

zakarumych:
Up

Lokathor:
I will try to have a look tonight.

Lokathor:
I'm going to accept this now, but I'm leaning against publishing it with the automatic build script feature selection.

I will think some more about this, but I think at least for now it will be a standard opt-in crate feature. It's semver compatible to make it automatic opt-in later, so I think initially we should take the "path of least commitment".

zakarumych:
What downsides you see in "automatic build script feature selection"?

PR #60, opened 2021-04-21
Set html_root_url attribute

tspiteri:
https://rust-lang.github.io/api-guidelines/documentation.html#crate-sets-html_root_url-attribute-c-html-root

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "bytemuck"
 description = "A crate for mucking around with piles of bytes."
-version = "1.5.2-alpha.0"
+version = "1.5.2-alpha.0" # remember to update html_root_url
 authors = ["Lokathor <zefria@gmail.com>"]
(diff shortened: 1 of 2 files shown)


Lokathor:
I asked @jyn514 and they said that this is apparently not a thing people should use in the common case, but the PR to update the docs about it has been stalled: https://github.com/rust-lang/api-guidelines/pull/230

tspiteri:
Ok. In my case apart from the docs.rs documentation (which uses extern-html-root-url) I keep a copy of the docs for the latest git default branch. The links were not working there, hence this PR. In the meantime I had updated my doc update script to use extern-html-root-url, so that solves the issue for me. So I think this PR can be closed.

jyn514:
> In the meantime I had updated my doc update script to use extern-html-root-url, so that solves the issue for me.

On nightly you can use `-Z rustdoc-map` which will do this automatically.

PR #63, opened 2021-06-04
Attach min_const_generics to a crate feature

chorman0773:
This allows use of the crate feature `min_const_generics` to implement `Zeroable` and `Pod` for arbitrarily sized arrays and adds a test to ensure the implementation operates.

It is noted that previously the feature was available behind detection logic in build.rs. This logic was left intact, however, it's possible that code may rely on the detected version without opting into the feature, and unknowning bind to MSRV 1.51, so I recommend it's removal.


Lokathor:
> it's possible that code may rely on the detected version without opting into the feature, and unknowning bind to MSRV 1.51, so I recommend it's removal.

I think I agree on that one. I didn't feel the best about having the auto-detection enabled. Since that `build.rs` detection stuff isn't published yet, can you adjust the PR to remove `build.rs` and just have all min-const-generic related support behind the cargo feature?

chorman0773:
That's done. It's only checking through the feature now.

Lokathor (approved):

PR #67, opened 2021-06-11
Allow building on spir-v

DJMcNab:
This allows this crate to be  used on the gpu side for [rust-gpu](https://github.com/EmbarkStudios/rust-gpu)

This removes a roadblock from using [crevice](https://crates.io/crates/crevice/) on the gpu side, and prevents confusing errors.

Complementary PR to https://github.com/EmbarkStudios/rust-gpu/pull/668

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,11 +100,19 @@ possibility code branch.
 /// Immediately panics.
 #[cold]
 #[inline(never)]
-fn something_went_wrong(src: &str, err: PodCastError) -> ! {
+fn something_went_wrong(_src: &str, _err: PodCastError) -> ! {
   // Note(Lokathor): Keeping the panic here makes the panic _formatting_ go
   // here too, which helps assembly readability and also helps keep down
   // the inline pressure.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -225,6 +233,10 @@ pub enum PodCastError {
 }
 impl core::fmt::Display for PodCastError {
   fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+    // Currently, this is unlikely to work for the spirv targets
+    #[cfg(target_arch = "spirv")]
+    write!(f, "PodCastError", self);
```
Lokathor, line 238: the `,self` part should be removed here. There's no interpolation positions within the literal.

DJMcNab: Hmm, I'm surprised that this compiled.

But yes, well spotted - fixed.

Lokathor:
I'll try to do a patch release later today.

Lokathor:
1.6.1 is out

DJMcNab:
Thanks for the swift release.

PR #69, opened 2021-06-12
Spir-v: just gate the Display impl instead

DJMcNab:
The old version didn't compile (because of the extra semicolon). 

It is extremely unlikely that anyone would ever want to use this impl on gpu, at least whilst the compiler is in its current state - if the impls did start working, they could just use the `Debug` impl.

This is not a breaking change, since all prior versions didn't compile on spir-v anyway. 

As a small note, there is no expectation for you to proactively support the spir-v target; beyond checking PRs which fixup any breakage - bytemuck stopping compiling would be the least breaking change in the entire rust-gpu ecosystem.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -231,12 +231,9 @@ pub enum PodCastError {
   /// exact.
   AlignmentMismatch,
 }
+#[cfg(not(target_arch = "spirv"))]
 impl core::fmt::Display for PodCastError {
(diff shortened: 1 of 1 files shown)


Lokathor:
reasonable, and this can be "undone" later if necessary by making the impl be everywhere with conditional compilation moving back inside.

DJMcNab:
Yeah, exactly. Moving the strange behaviour to compile time is useful.

Lokathor:
released as 1.6.3

PR #73, opened 2021-08-11
Support WASM SIMD's `v128` type

CryZe:
Due to the MSRV it is behind a feature gate.


Lokathor:
once again you're doing Ferris's work.

Can you put on the doc attribute thing so that the generated docs show that a feature is involved? Like `safe_arch` has all over.

Lokathor:
in the newest rustdoc the attribute won't be required, it will automatically be generated.

PR #80, opened 2021-12-21
Implement `Pod` and `Zeroable` for AArch64 NEON Vectors

CryZe:
The NEON intrinsics will be stabilized in Rust 1.59 (and already are on nightly).


Lokathor:
ohhh i think i broke the CI when i added portable simd.

uhm, just remove that --all-features test on beta.

Lokathor (approved):

PR #83, opened 2021-12-30
Make `Pod` and `Zeroable` derives work for some generic structs and improve docs

### LINKED ISSUE #51: Documentation for `derive` feature.
Currently, there is no documentation for the `derive` feature, which allows for easy implementations for `Pod` and `Zeroable`, which makes it difficult for people to effectively use the crate, especially as the two derive macros show up in the documentation.

Suggested fixes:

* Provide a note in the documentation suggesting use of the `derive` feature to get `derive` implementations for `Pod` and `Zeroable` (provide code example)
* Make the `derive` macro a default feature.

### LINKED ISSUE #75: `derive(Pod)` should work for `repr(c)` structs with generics fields where all fields have the same type
According to [this](https://doc.rust-lang.org/reference/type-layout.html#reprc-structs) definition of `repr(C)` in Rust, structs where all fields have the same type are laid out like arrays (no padding anywhere). The reference doesn't directly state this, but it follows from the given layout algorithm (unless I'm wrong).

LukasKalbertodt:
Fixes #75 
Fixes #51 

- `derive(Zeroable)` now works for structs with generic fields. The `impl` is bounded with `T: Zeroable` for each struct field `T`.
- `derive(Pod)` now works for structs where all fields have the same generic type. Again, the `impl` is bounded with `T: Pod` then.
- I improved some documentation and mentioned the derive feature in the crate-level docs.

**Note**: the code emitted by the derive macros is slightly different with this PR. This has the unfortunate consequence that some compiler errors (for when a struct's field does not implement `Pod`) are worse than before. However, this was due to a Rust issue which has been fixed and will be released on January 13 as 1.58. See #75 for more details.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -42,14 +42,12 @@ pub fn derive_pod(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
   proc_macro::TokenStream::from(expanded)
 }
 
-/// Derive the `Zeroable` trait for a struct
+/// Derive the `Zeroable` trait for a struct.
 ///
-/// The macro ensures that the struct follows all the the safety requirements
-/// for the `Zeroable` trait.
-///
-/// The following constraints need to be satisfied for the macro to succeed
-///
-/// - All fields ind the struct must to implement `Zeroable`
+/// The macro ensures that the struct follows all the safety requirements for
+/// the `Zeroable` trait. If your struct has generic fields, the generated
+/// `impl` will have a `T: Zeroable` bound for each generic type `T` of your
+/// struct.
 ///
 /// ## Example
 ///
@@ -160,7 +158,7 @@ fn derive_marker_trait_inner<Trait: Derivable>(
 ) -> Result<TokenStream, &'static str> {
   let name = &input.ident;
 
-  let (impl_generics, ty_generics, where_clause) =
+  let (impl_generics, ty_generics, original_where_clause) =
     input.generics.split_for_impl();
 
   let trait_ = Trait::ident();
@@ -168,6 +166,13 @@ fn derive_marker_trait_inner<Trait: Derivable>(
   let asserts = Trait::struct_asserts(&input)?;
   let trait_params = Trait::generic_params(&input)?;
   let trait_impl = Trait::trait_impl(&input)?;
+  let extra_bounds = Trait::extra_where_bounds(&input)?;
+
+  // Merge existing bounds with our bounds.
+  let mut where_clause = original_where_clause
+    .map(|w| if w.predicates.trailing_punct() { quote! { #w } } else { quote! { #w, }})
+    .unwrap_or(quote! { where });
+  where_clause.extend(extra_bounds);
 
   Ok(quote! {
     #asserts
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -13,14 +13,13 @@ use crate::traits::{Contiguous, Derivable, Pod, TransparentWrapper, Zeroable};
 /// Derive the `Pod` trait for a struct
 ///
 /// The macro ensures that the struct follows all the the safety requirements
-/// for the `Pod` trait.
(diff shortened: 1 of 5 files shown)


Lokathor:
I thought I had replied last week but I guess it didn't post, very sorry.

I like what I see but I literally don't use proc-macros at all so I'd like at least one other person to have a look at this and check that things seem right.

5225225:
The error message for when the types are different looks to be wrong now. Because the derive *can* be done for generic types, just not ones that have different types.

derive/src/lib.rs
```diff
@@ -42,14 +41,12 @@ pub fn derive_pod(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
-/// The macro ensures that the struct follows all the the safety requirements
-/// for the `Zeroable` trait.
-///
-/// The following constraints need to be satisfied for the macro to succeed
-///
-/// - All fields ind the struct must to implement `Zeroable`
+/// The macro ensures that the struct follows all the safety requirements for
+/// the `Zeroable` trait. If your struct has generic fields, the generated
+/// `impl` will have a `T: Zeroable` bound for each generic type `T` of your
+/// struct.
```
Nemo157, line 49: This says it will add a `T: Zeroable` bound, but for a struct like `struct Foo<T: Iterator> { foo: T::Item }` it looks like it will actually generate a `T::Item: Zeroable` bound (which is more correct, but not what derives generally do).

danielhenrymantilla: This. I personally prefer what this derive does, but it's important that whatever choice is made, it's done being aware that it's not the stdlib's approach (cc @Lokathor).

I personally prefer the approach here, because the stdlib one is what makes:

```rust
#[derive(PartialEq, Eq)]
struct Ptr<T>(*const T);
```

only yield comparable pointers when the pointees are.

Lokathor: This is a tough one.
* I'd generally prefer that the *derives* be "extra picky" because this is an unsafe trait after all, and you *can* always just implement manually if you think you know better than the derive's reasoning.
* I do agree that the standard library is a little bit wrong on this one, and I've been bit before by a standard library derive being too conservative (when I wrote `VolAddress<T,R,W>`).

However, I think that since this is becoming an increasingly used crate **we should have the derives work like the stdlib derives**. Don't want people to have any sudden surprises happening.

danielhenrymantilla: The advantage of the default approach will be that we'll have correctly-spanned error messages even in older rust versions 🙂

danielhenrymantilla (commented):
Nice initiative 🙂 

  - I'd not recommend enabling `syn`'s extra-traits, at least not for a syntactical type comparison which is nonetheless brittle;
  - I'd thus delegate to a static assertion that the types are equal (if you still want the nice error message, then you can use `.to_string()` to perform comparison without `extra-traits`, but there should be a const assertion that the types are nonetheless equal).
  - Regarding `Pod`, it could be nice to also support `repr(C)` / `repr(transparent)` wrappers, as well as `repr(packed)` structs (for a follow-up PR, I imagine).
      - Not necessarily a good thing, but it is possible to soundly implement `Pod` for a fully generic struct, by relying on post-monomorphization errors: those are better than panics, but worse than classic trait const assertions. Indeed, if the code "links" (performs codegen) correctly, then it won't panic; but it can still fail to codegen despite `cargo check` passing.
        If you are interested, I got a proof-of-concept for an `AsBytes` trait (which is the part that `Pod` is checking, there), in [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6e531ee532b6badffbdd2322a54aa9a9).

LukasKalbertodt:
Sorry for the late reply! Unfortunately, I don't need this change anymore and don't really have the energy to work on this right now. Since this PR would require a non-trivial rebase and has several unresolved problems, I will simply close it now. I still think the general direction of this PR is good and that the derives should be generalized. But yeah, closing, sorry!

PR #88, opened 2022-01-14
nightly_no_coverage

MaulingMonkey:
⚠️ I don't know if this should actually be merged as is. ⚠️

This relies on `#![feature(no_coverage)]` in the root of **each crate using** the derive macros, which is awkward, because it's quite possible only one crate opted into the feature.  Perhaps it'd be better to generate `#[cfg_attr(feature = "nightly_no_coverage", no_coverage)]` instead, and then have the crates that want `#[no_coverage]` define a feature on themselves?

rust-lang tracking issue: https://github.com/rust-lang/rust/issues/84605

## What is this?

Unit test coverage tools are nice, but the never-called anonymous functions that bytemuck uses to assert things are safe cause [grcov](https://github.com/mozilla/grcov) and other tools to scold us for not testing said functions:

image

This changelist fixes that:

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,11 +26,12 @@ unsound_ptr_pod_impl = []
 
 # NOT SEMVER SUPPORTED! TEMPORARY ONLY!
 nightly_portable_simd = []
+nightly_no_coverage = ["bytemuck_derive/nightly_no_coverage"]
 
 [dependencies]
 # use the upper line for testing against bytemuck_derive changes, if any
-#bytemuck_derive = { version = "1.0.1-alpha.0", path = "derive", optional = true }
-bytemuck_derive = { version = "1", optional = true }
+bytemuck_derive = { version = "1.0.2-alpha.0", path = "derive", optional = true }
+#bytemuck_derive = { version = "1.0.2-alpha.0", optional = true }
 
 [package.metadata.docs.rs]
 # Note(Lokathor): Don't use all-feautures or it would use `unsound_ptr_pod_impl` too.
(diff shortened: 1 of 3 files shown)


MaulingMonkey:
An alternative suggestion: instead of *sometimes* marking these uncalled check fns as `#[no_coverage]`, instead *always* mark them as:
```rust
#[cfg(not(unsafe_unsound_unstable_disable_bytemuck_soundness_checks))]
```
Which could then be used ala:
```sh
export RUSTFLAGS="--cfg unsafe_unsound_unstable_disable_bytemuck_soundness_checks -Zinstrument-coverage"
export LLVM_PROFILE_FILE="profraw/profraw-%%p-%%m.profraw"
cargo test --target-dir=target/coverage
grcov . -s . --binary-path target/coverage/debug -t html --branch --ignore-not-existing -o grcov
```
If `#[no_coverage]` ever stabilizes, and bytemuck's MSRV bumps to include it, the cfg could then be turned into a noop in favor of `#[no_coverage]`.

Is that more reasonable?  Name workshopping / suggestions for the `--cfg` name are welcome.

MaulingMonkey:
(Perhaps the name of such a cfg should avoid being `bytemuck` specific, to allow any other crates with similar static assertion fns to reuse the name and similarly opt out... `unsafe_unsound_unstable_remove_static_asserts_for_coverage`?)

Lokathor:
Well I'm naturally dubious of a "disable soundness checks" thing, but i guess if it's opt-in.

MaulingMonkey:
(Closed in favor of https://github.com/Lokathor/bytemuck/pull/89)

Issue #85, opened 2022-01-02
Does `Pod` have to require `Zeroable`?


repi:
There probably is a reason or specific history why `Pod` derives `Zeroable`, and I may be missing something but couldn't find a motivation in the docs or code. 

Logically to me it feels like all _zeroable_ types would be _pod_ types, but all _pod_ types are not necessarily _zeroable_. Does that make sense? 

Been experimenting with a bunch of use cases in our code (also see #84) and while I can see a use for _zeroable_ types, it is (for us at least) very rare and we do have a lot of _pod_ types that due to this requirement also have to derive and be _zeroable_ and expose `::zereod()` to the users without it really making much sense to every crate a zeroable implementation of them - so creates some extra noise and potential confusion with this requirement (for us).

Lokathor:
If the type is Pod, that means you're allowed to cast `[u8; N]` into that type, where N is however many bytes that type is.

Supposing that the array contains all 0 values, then you'd have made the zeroed instance of that type.

Thus, while it might not be useful to have a zeroed instance of the type, it must be legal for that instance to exist. If it's not legal, then the type shouldn't be Pod in the first place.

termhn:
Seems to me that perhaps some confusion is the `Zeroable` does not require for `::zeroed` to make *logical sense*, but rather just that `zeroed` cannot cause the type to expose unsafe behavior, which are quite different requirements. Though perhaps @repi already realizes that and I'm missing something, haha. This does mean that having `zeroed` method exposed when deriving `Zeroable` is a bit inconvenient if we don't actually want users readily creating the zeroed version. But, the user would have to import the `Zeroable` trait to access it anyway, so I don't think it's a huge deal.

Issue #84, opened 2022-01-02
Support for pod structs with enums in them?


repi:
Currently it looks like pod/zeroable structs do not support having enum fields in them, is this correct?

Such as:

```rust
#[derive(Copy, Clone, Pod, Zeroable)]
#[repr(C)]
struct TestWithEnum {
  a: u16,
  b: u16,
  e: ContiguousEnum,
}

#[repr(u8)]
#[derive(Clone, Copy, Contiguous)]
enum ContiguousEnum {
  A = 0,
  B = 1,
  C = 2,
}
```

And the derive macros for `Pod` and `Zeroable` doesn't support enums.

Though enums that use a primitive representation could be considered pod types (even without the `Contiguous` additional requirement`), and if such an enum implements a variant with value `0` it could support implement `Zeroable` trait as well right?

Is there some additional limitations here that I'm not aware of or has it simply not been implemented yet?

We have a bunch of use cases in native and shader code where we have pod structs (that do have enums in) that we want to use bytemuck on for the additional compile-time verification that they use the right representation and limitations, and these structs we cast back and forth to u8 slices.

Lokathor:
> Is this intended as the restriction of why enums are not supported also

Yes.

Here's how things could go wrong:

```rust
enum MyEnum {
  Whatever = 0,
  AnotherTag = 1,
}
// this is unsound
unsafe impl Pod for MyEnum {}

let mut my_enum = MyEnum::Whatever;
let x: &mut u8 = cast_mut(&mut my_enum);
*x = u8::MAX; // now `my_enum` holds an illegal bit pattern
match my_enum {
  // this match expression is gonna go sideways
}
```

Still, many enums can correctly be `Zeroable`.

Lokathor:
As a side note: your `TestWithEnum` struct will actually have a padding byte as the 6th byte, which I hope the Pod derive would detect and crash on, even if it did otherwise accept the enum.

CryZe:
Almost sounds like it might make sense to have a read only variant of Pod that enums could implement.

Lokathor:
You're not the first to propose such a thing.

When I was first developing `bytemuck` I asked in the Discord and everyone considered the ability to convert `&T` to `&[u8]` a "not very useful" thing on its own, so I didn't make a separate trait for it.

At this point, it would be a breaking change to insert into the existing `Pod: Zeroable` trait tree, but just as a totally separate trait would be fine. We could even have it blanket impl for all Pod types.

[...]

Issue #41, opened 2020-10-02
Split `Pod` into two traits


zakarumych:
Traits `NoPadding` and `AnyBitPattern` that would allow "casting" like this

```
fn cast<T, U>(T) -> U
where
    T: NoPadding,
    U: AnyBitPattern,
{
   ...
}
```

Then `Pod` is simply both.

zakarumych:
As a simple use-case, this will enable us to reinterpret structures with `bool` as bytes.

Lokathor:
When the 1.0 of this crate was being developed, no one considered the ability to view a `&bool` as a `&u8` very valuable, so this particular trait split was skipped on.

I think this would be a breaking change to fit in to the existing crate.

Because of the safe transmute project, any work on a 2.0 is being held off until that project is done and stabilized. However, at that time we might be able to split up the traits like this.

zakarumych:
Would it be a breaking change though?
Relaxing requirements on functions doesn't seem to break anything.

If we add blanket impls like this:

```
impl<T: Pod> NoPadding {}
impl<T: Pod> AnyBitPattern {}
```

instead of the other way around then all existing implementation would still work.

Lokathor:
I admit I'm not a trait expert, so with careful design it might be done I suppose.

I'll point some of the trait folks I know at this topic and maybe they can weigh in.

zakarumych:
Once again I'd really like this feature, as I want to cast bytes to other types which can have padding but accept any bit pattern.

Lokathor:
We could have a NoPadding trait that lets you turn &A into &[u8], and just have it as a separate thing from Pod. That wouldn't be a breaking change.

zakarumych:
And currently I need the reverse. Turning `&[u8]` into `A: AnyBitPattern`.
Although `NoPadding` has its use cases as well.
Do you still think proposed solution above may be a breaking change?

Lokathor:
Yeah. No one I spoke to thought that it was possible to insert and extra trait into the trait tree.

zakarumych:
I'm curious. Did they provide a counterexample?

Lokathor:
No one showed that it couldn't be done, but no one definitely showed it can be done with absolutely no chance of breaks.

I'd rather just have a separate trait, which is assured to not break any existing code.

zakarumych:
Do you mean that you want separate functions as well?

Lokathor:
[...]

PR #91, opened 2022-02-07
Add `NoPadding`, `AnyBitPattern`, and `CheckedBitPattern` traits

fu5ha:
This PR adds three new traits, `NoPadding`, `AnyBitPattern`, and `CheckedBitPattern`.

## Motivation

Similar to #41 and motivated for the same reasons as #84 and #85. Specifically, the motivations for this PR are:

1. to allow casting by move and by *immutable* reference *from* any type that has no padding, *into* a `Pod` type.
2. to allow casting *to* types that have invalid bit patterns by checking whether the beginning data violates the relevant rules.

Most notably, this allows `char`, `bool`, and C-style enums.

## Overview

`AnyBitPattern` has all the requirements of `Pod` except it does not require the type to have no padding -- most relevantly, it requires any bit pattern of the correct size to be a valid instance of `Self`.

`NoPadding` has all the requirements of `Pod` except it does not require `AnyBitPattern`--i.e. types that have illegal bit patterns can still implement it.

These allow casting by move and by *immutable* reference *from* any type that is `NoPadding` *into* an `AnyBitPattern` type. *Mutable* reference casting is disallowed *from* `NoPadding` because the original data could be modified in an illegal way, and is disallowed *to* `AnyBitPattern` types as there could be padding bytes that would become uninitialized on write -- unsound for the original `NoPadding` type. So, mutable reference casting still requires full `Pod` on both sides.

`CheckedBitPattern` is a trait which allows a checked version of casting functions for types that are not `AnyBitPattern`. The way this works is that we define a `Bits` ty which has the same layout as `Self` except it has no invalid bit patterns -- and therefore can implement `AnyBitPattern`. We then define a function that takes an `&Bits` and returns whether or not that instance of the `Bits` type contains any invalid bit patterns for `Self`. If not, we can go ahead and cast as though `Self` was itself `AnyBitPattern`. This allows move and *immutable* ref casting *from* a `NoPadding` type *into* a `CheckedBitPattern` type, and `mutable` ref casting *from* a `Pod` type *into* a `CheckedBitPattern + NoPadding` type.

## Implementation

Moved the business logic of all the casting functions that were previously in `lib.rs` into a new module, `internal`, and relaxed the trait bounds to only be `Copy`. This makes them unsafe. Then, just call those from the `lib.rs` functions, increasing the trait bounds to make them safe (and fully equivalent to before).

~~Added `relaxed` module, which is entirely equivalent to the functions in `lib.rs` other than relaxing the `Pod` requirement to `NoPadding` in cases where possible. Maybe should add the `AnyBitPattern` compliment trait to `NoPadding` as well, like @zakarumych suggested in #41? That would slightly relax the requirements specifically on the by-move `cast` and `try_cast` functions. Maybe worth it to just do that now so it won't be breaking if someone ends up wanting it in the future...~~

Relaxed the bounds of functions in `lib.rs` to use `NoPadding` and `AnyBitPattern` where possible/relevant. This is not breaking as there is a full blanket impl of `NoPadding` and `AnyBitPattern` for any `T: Pod`.

Added `checked` module, which further relaxes requirements from `relaxed` module to allow casting *to* `CheckedBitPattern` types where possible, as described above.


termhn:
Okay. Reverted everything for the existing functions and added new versions... Updated original post's implementation section describing what I did.

I'm not convinced that `TryFrom` is able to do everything that I need for use case 2, or, if it did, it would basically require re-implementing the whole of bytemuck in and external crate in order to do so without violating orphan requirements... and even then I think it wouldn't be as simple/elegant of a solution.

Lokathor:
It's absolutely not a perfect solution for all cases, but in many cases you can write `impl TryFrom<&[u8]> for &MyType` and do any checks based on your fields and get on with it. And an interested party could make a proc-macro for all this. As the complexity built you'd eventually end up with serde, but instead of getting back an owned value you'd get back a reference. It's not as popular because often people don't want to re-interpret in place to begin with, but it's possible.

The rest of this I'll try to have a look at tomorrow.

Initial Bikeshed: I don't think that "MaybePod" is the right name for this checked conversion system. Not sure what the right name is, but it feels like some other name would center the idea better.

zakarumych:
If you going to add `NoPadding` trait, please add dual `AnyBitPattern` trait as well.

When I asked for this originally, my proposal was rejected. So I just ended up doing it manually where needed, I didn't want to stuck in lengthy debate.

I still believe that relaxing requirements for functions with implementation of new traits for types that already implement `Pod` would not be a breaking change.

`MaybePod` trait name looks misleading. If I understand this correctly. Deriving `MaybePod` should generated `Pod` structure with same layout and a function to attempt conversion from generated `Pod` struct to `MaybePod` type. So it has nothing to do with `Pod`ness of `MyabePod` type. Maybe call it `TryFromPod` instead?

zakarumych:
`MaybePod: NoPadding` bound look unnecessary.
Casting from `T` to `T::PodTy` must require `NoPadding`. Casting from `T::PodTy` to `T` - doesn't

termhn:
> 



> `MaybePod: NoPadding` bound looks unnecessary. Casting from `T` to `T::PodTy` must require `NoPadding`. Casting from `T::PodTy` to `T` - doesn't

For by-move casting, true, however not requiring this bound makes deriving `PodTy` much harder. Also, it is required if we want to cast `&T::PodTy` to `&T` (or `&mut`, also casting slices).

termhn:
> If you going to add `NoPadding` trait, please add dual `AnyBitPattern` trait as well.

Okay, it shouldn't be too hard to add, will work on that today.

> I still believe that relaxing requirements for functions with implementation of new traits for types that already implement `Pod` would not be a breaking change.

Me too, but 🤷‍♀️ 

> `MaybePod` trait name looks misleading. If I understand this correctly. Deriving `MaybePod` should generated `Pod` structure with same layout and a function to attempt conversion from generated `Pod` struct to `MaybePod` type. So it has nothing to do with `Pod`ness of `MyabePod` type. Maybe call it `TryFromPod` instead?

There's one difference from what you describe which is that the function doesn't convert from `T::PodTy` to `T`, it *validates* whether a reinterpret cast (i.e. what's done by the rest of the casting functions in the crate) is okay from `T::PodTy` to `T`. I actually originally called it `MaybeCastFromPod` when I was developing it

Some other naming ideas:

`CheckedPod`
`CheckedBitPattern`
`MaybeCastFromPod`

zakarumych:
>  it is required if we want to cast `&T::PodTy` to `&T`

as far as I understand it - no.

Having
```rust
#[repr(C)]
struct Foo {
  a: u8,
  b: u16,
}
```
with padding byte between `a` and `b` fields, it should be safe to cast `&[u8; 4]` to `&Foo`, even though second byte won't be relevant to `Foo` value.

zakarumych:
To me `T: MaybePod` sounds like it tells at runtime if `T` is actually `Pod` type.

termhn:
> > it is required if we want to cast `&T::PodTy` to `&T`
> 
> as far as I understand it - no.
> 
> Having
> 
> ```rust
> #[repr(C)]
> struct Foo {
>   a: u8,
>   b: u16,
> }
> ```
> 
> with padding byte between `a` and `b` fields, it should be safe to cast `&[u8; 4]` to `&Foo`, even though second byte won't be relevant to `Foo` value.

No, because now that you've viewed those bytes as a `&Foo` any bytes that are padding in `Foo` must now be treated as uninitialized--yet you may still have view of them as valid `u8` in other references to the original slice. Even with `&mut` this is the same because after the mutable reference goes out of scope, you still have wherever the original mutable byte slice came from that is (or at least might be) assuming those bytes are initialized.

termhn:
> To me `T: MaybePod` sounds like it tells at runtime if `T` is actually `Pod` type.

That's... sorta the idea? except it's telling if a certain instance of T can be counted as Pod. It's not a perfect analogy, admittedly

termhn:
> No, because now that you've viewed those bytes as a `&Foo` any bytes that are padding in `Foo` must now be treated as uninitialized--yet you may still have view of them as valid `u8` in other references to the original slice. Even with `&mut` this is the same because after the mutable reference goes out of scope, you still have wherever the original mutable byte slice came from that is (or at least might be) assuming those bytes are initialized.

FWIW, this is also why `AnyBitPattern` is only applicable to `cast`, `try_cast`, and owned `alloc`-backed things like `Vec`/`Box`

zakarumych:
> > To me T: MaybePod sounds like it tells at runtime if T is actually Pod type.

> That's... sorta the idea? except it's telling if a certain instance of `T` can be counted as `Pod`. It's not a perfect analogy, admittedly

That's exactly the other way around, isn't it? It's telling if certain instance of `Pod` can be  counted as `T` :)

zakarumych:
> any bytes that are padding in `Foo` must now be treated as uninitialized

Do they?

termhn:
> Do they? I was under impression this only applies for `&mut Foo`. Because you'd be able to assign uninit bytes with `*foo = Foo { ... }`

I guess we should get someone else to verify, but I'm 90% sure that it's not allowed even under shared ref. I could be wrong though.

termhn:
Alright, according to miri, you are indeed correct and the shared ref case is fine, though the mutable case is not. I guess we could bound the mutable case functions as being `MaybePod + NoPadding` instead of making it (or whatever `MaybePod` ends up getting called) always bounded on `NoPadding`.

Lokathor:
So, just to add a slightly longer follow up to the above point about data re-viewing: if you have a `&[u8]` and view it as `&TypeWithPaddingInside` it doesn't *actually* make any of the bytes become uninitialized. Only an actual write of the memory will cause bytes to become either initialized or un-initialized. If you convert `&mut [u8]` to `&mut TypeWithPaddingInside` and then write your whole struct to that unique reference then the bytes that are padding within the struct become un-initialized. However, if you write a particular field within the struct that doesn't have padding (such as a single `i32` field) then that only writes that one part of the memory and doesn't de-initialize any memory. So shared reference re-views are, in general, much much easier to be done safely because you can't de-initialize any memory.

termhn:
> I suppose casting `&mut MaybeUninit<T::PodTy>` to `&mut T` can also be permitted without `NoPadding` bound. If it is actually initialized and call to `MaybePod::cast_is_valid` returns true ofc.

True but I don't think this is particularly useful :P

Also, I've added AnyBitPattern, moved the more relaxed functions into `lib.rs` again after discussions in Discord\*, and reworked `MaybePod` into `CheckedBitPattern`. It now is not a subtrait of `NoPadding`, so I added that bound to the mut functions where needed.

\*note: which confirmed that as long as there are true blanket impls of `NoPadding` and `AnyBitPattern` for any `T: Pod` (which there are), this is not a breaking change.

src/lib.rs
```diff
@@ -150,14 +157,7 @@ pub fn bytes_of<T: Pod>(t: &T) -> &[u8] {
 /// empty slice might not match the pointer value of the input reference.
 #[inline]
 pub fn bytes_of_mut<T: Pod>(t: &mut T) -> &mut [u8] {
```
zakarumych, line 159: If you replace `T: Pod` with `T: NoPadding + AnyBitPattern`, then any `Pod` type would be valid, but also other types that (at least in generic context) are known to implement `NoPadding` and `AnyBitPattern`

Lokathor:
One thing that I realized suddenly is that the NoPadding rules must disallow UnsafeCell

src/lib.rs
zakarumych: I guess I'm in the wrong timezone to participate in discord discussions.

My point is that `T: Pod` is unnecessary strict compared to `T: NoPadding + AnyBitPattern`.
I can imagine some generic context where generic type has `NoPadding + AnyBitPattern` bound.
If that code is generated by proc-macro it may be hard to replace with `Pod`.
I don't have real case of this naturally. But we can play it safe.

termhn: Sorry for the delay, but I can agree with this. Changed in d5abea1

termhn:
Alright I think this is now ready for "final review"

src/checked.rs
```diff
@@ -0,0 +1,419 @@
...
+/// [`is_valid_bit_pattern`]: CheckedBitPattern::is_valid_bit_pattern
+/// [`Pod`]: crate::Pod
+pub unsafe trait CheckedBitPattern: Copy {
+  /// `Self` *must* have the same layout as the specified `Bits` except for
+  /// the possible invalid bit patterns being checked during [`is_valid_bit_pattern`].
+  ///
+  /// [`is_valid_bit_pattern`]: CheckedBitPattern::is_valid_bit_pattern
+  type Bits: AnyBitPattern;
+
+  /// If this function returns true, then it must be valid to reinterpret `pod` as `&Self`.
```
JakobDegen, line 122:
```suggestion
  /// If this function returns true, then it must be valid to reinterpret `bits` as `&Self`.
```

src/checked.rs
```diff
@@ -0,0 +1,419 @@
...
+///
+/// ## Failure
+///
+/// * If the types don't have the same size this fails.
+/// * If `a` contains an invalid bit pattern for `B` this fails.
+#[inline]
+pub fn try_cast<A: NoPadding, B: CheckedBitPattern>(
+  a: A,
+) -> Result<B, CheckedCastError> {
+  let pod = unsafe { internal::try_cast(a) }?;
```
JakobDegen, line 235:
```suggestion
  let pod = unsafe { internal::try_cast(a)? };
```
I have no reason to prefer either form, but the functions above do it one way and the functions below do it a different way

JakobDegen (commented):
Just nits, looks good from a soundness pov

src/anybitpattern.rs
```diff
@@ -0,0 +1,43 @@
...
+/// This limits what you can do with a type of this kind, but also broadens the
+/// included types to `repr(C)` structs that contain padding. Notably, you can only cast
+/// *immutable* references and *owned* values into [`AnyBitPattern`] types, not
+/// *mutable* references.
+///
+/// [`Pod`] is a superset of [`AnyBitPattern`], meaning that any `T: Pod` is also
+/// [`AnyBitPattern`] but any `T: AnyBitPattern` is not necessarily [`Pod`].
+///
+/// [`AnyBitPattern`] is a subset of [`Zeroable`], meaning that any `T: AnyBitPattern`
+/// is also [`Zeroable`], but any `T: Zeroable` is not necessarily [`AnyBitPattern  ]
```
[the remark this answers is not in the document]
zakarumych: No, the above is backwards, and this one is correct.
`AnyBitPattern` types set is subset of `Zeroable` types set.

src/anybitpattern.rs
```diff
@@ -0,0 +1,43 @@
...
+///
+/// The requirements for this is very similar to [`Pod`],
+/// except that it doesn't require that the type contains no padding bytes.
+/// This limits what you can do with a type of this kind, but also broadens the
+/// included types to `repr(C)` structs that contain padding. Notably, you can only cast
+/// *immutable* references and *owned* values into [`AnyBitPattern`] types, not
+/// *mutable* references.
+///
+/// [`Pod`] is a superset of [`AnyBitPattern`], meaning that any `T: Pod` is also
+/// [`AnyBitPattern`] but any `T: AnyBitPattern` is not necessarily [`Pod`].
```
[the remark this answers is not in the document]
zakarumych: Right, see below, `AnyBitPattern` to `Zeroable` relation has exactly same explanation, but is called `subset` (correctly so).

src/checked.rs
zakarumych: I'd say "Keep least amount of ops in the `unsafe` block".
Therefore `?` should be outside unsafe block, unless unwrapped value is used in the unsafe block later. Which is not the case here.

termhn:
Thanks for the review @JakobDegen and @zakarumych :)

fu5ha:
Hey @Lokathor friendly ping, do you have more thoughts on this?

fu5ha:
I think there's an issue with incremental compilation for CI, as the issues it's complaining about don't make sense -- the line it's pointing at in `lib.rs` does not exist x)

fu5ha:
Not sure how (or if I can) do a `cargo clean` or equivalent on CI

fu5ha:
🥳 tests pass

PR #94, opened 2022-03-29
Support unions in derive macros

fu5ha:
This adds support for unions in the derive macros.

The logic goes:

- All unions are `AnyBitPattern` because actually accessing them is unsafe anyhow
- ~~Unions are `NoPadding` (and `Pod`, since they're always `AnyBitPattern`) if~~
    - ~~All their fields are `NoPadding`~~
    - ~~All their fields are the same size (therefore no padding added at the end of any variants)~~

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -5,8 +5,8 @@ extern crate proc_macro;
 mod traits;
 
 use proc_macro2::TokenStream;
-use quote::quote;
-use syn::{parse_macro_input, DeriveInput};
+use quote::{quote, quote_spanned};
+use syn::{parse_macro_input, DeriveInput, spanned::Spanned};
 
 use crate::traits::{
   AnyBitPattern, Contiguous, Derivable, CheckedBitPattern, NoPadding, Pod, TransparentWrapper, Zeroable,
@@ -223,8 +223,9 @@ pub fn derive_contiguous(
 
 /// Basic wrapper for error handling
 fn derive_marker_trait<Trait: Derivable>(input: DeriveInput) -> TokenStream {
+  let span = input.span();
   derive_marker_trait_inner::<Trait>(input).unwrap_or_else(|err| {
-    quote! {
+    quote_spanned! { span =>
       compile_error!(#err);
     }
   })
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -1,9 +1,9 @@
 use proc_macro2::{Ident, TokenStream, TokenTree};
-use quote::{quote, quote_spanned, ToTokens};
+use quote::{quote, quote_spanned, ToTokens, format_ident};
 use syn::{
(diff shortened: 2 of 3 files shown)


JakobDegen:
>     * Unions are `NoPadding` (and `Pod`, since they're always `AnyBitPattern`) if
>       
>       * All their fields are `NoPadding`
>       * All their fields are the same size (therefore no padding added at the end of any variants)

This is probably not correct. The following code passes miri:
```rust
fn main() {
    union U {
        _a: bool,
    }
    unsafe {
        let _u: U = core::mem::MaybeUninit::uninit().assume_init();   
    }
}
```

The current stance of [UCG](https://github.com/rust-lang/unsafe-code-guidelines/issues/73) is somewhat undecided, but the issue seems to lean towards "all data is accepted for all unions." Of course, it still requires unsafe code to get the `uninit` bytes into the union, but still I don't think this impl should exist.

fu5ha:
:( That is... quite inconvenient

> Of course, it still requires unsafe code to get the `uninit` bytes into the union, but still I don't think this impl should exist.

My view is that if you use unsafe to do this sort of thing then all bets are off, so I still think that the impl is fine

JakobDegen:
I don't think that's the case at all, because such unsafe code can be behind a sound API.

```rust
// Library crate A
#[derive(NoPadding)]
union U {
    _a: bool,
    b: u8,
}

// Library crate B
pub fn lib_fn(x: i32) -> (bool, U) {
    if x == 0 {
        (true, U { b: '0' })
    } else {
        // In this case we don't care what `U` is
        (false, MaybeUninit::uninit().assume_init()
    }
}

// Upstream
fn main() {
    let (_, u) = lib_fn(1);
    dbg!(bytemuck::transmute::<_, [u8; 1]>(u))
}
```

The only unsafe code here is in library crate B, but I don't see how you can argue that it's unsound

Lokathor:
I think we need to go with JakobDegen on this one. All unions are effectively *entirely* padding, in a sense, at least for the purposes of our traits. Since all bytes of a union can be uninit, we can't have them be `NoPadding`.

fu5ha:
hm, I can buy that I think. I will remove the impls that depend on that from this PR and rebase on main

Lokathor:
It does hint that we may want to rename `NoPadding` to `AllBytesInit`, since it's not released yet. I'll leave that call up to you since you've lead the design on this one.

fu5ha:
Hmmmmmmm

fu5ha:
I dunno, I still think `NoPadding` is the more elegant name here, and still works well, if slightly less precise? The union case seems like a bit of unique side case. I would say we stay with `NoPadding` for now, and we could re-evaluate when the time for 2.0 comes whether we want to keep it or change.

JakobDegen:
Maybe `NoUninit`?

Lokathor (approved):

PR #106, opened 2022-05-18
Note on shared references to Pod-like types

HeroicKatora:
We need to say why `Cell<u8>: Pod` etc. is _not_ allowed.


Lokathor:
After "The type must have no interior mutability.", it might be best to add "This means no UnsafeCell, or any type that contains an UnsafeCell."

Separately, uf we're going to document what types definitely don't fit, then pointers should also get a mention in the docs as something people should never use with this crate.

PR #111, opened 2022-06-25
feat: add TransparentWrapperAlloc trait

### LINKED ISSUE #110: TransparentWrapper and alloc
Is it possible to have something like `cast_vec` for `TransparentWrapper` types?

```rs
#[repr(transparent)]
#[derive(TransparentWrapper)]
struct Bar(Foo);

let a: Vec<Foo> = vec![Foo::new()];
let b: Vec<Bar> = Bar::wrap_vec(a);
let c: Vec<Foo> = Bar::peel_vec(b);
```

I'm not sure if this is sound but I cannot see any reason why not. I can imagine this being implemented either on the `TransparentWrapper` trait itself using feature gates, or using an extension trait that appears inside the `allocation` module.

TheEdward162:
closes #110

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -306,20 +306,21 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>: TransparentWrapper<Inner> {
     Self: Sized,
     Inner: Sized
   {
-    // Not sure if these are needed
-    assert!(size_of::<*mut Inner>() == size_of::<*mut Self>());
-    assert!(align_of::<*mut Inner>() == align_of::<*mut Self>());
-
     let mut s = core::mem::ManuallyDrop::new(s);
+
+    let length = s.len();
+    let capacity = s.capacity();
+    let ptr = s.as_mut_ptr();
+
     unsafe {
       // SAFETY:
       // * ptr comes from Vec (and will not be double-dropped)
       // * the two types have the identical representation
       // * the len and capacity fields are valid
       Vec::from_raw_parts(
-        s.as_mut_ptr() as *mut Self,
-        s.len(),
-        s.capacity()
+        ptr as *mut Self,
+        length,
+        capacity
       )
     }
   }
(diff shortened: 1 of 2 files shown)


Lokathor (approved):
This looks good. I also posted it on the discord for people to have a look at, and if there's no complaints from them I'll merge it later today (or maybe tomorrow).

src/allocation.rs
```diff
@@ -297,3 +297,55 @@ pub fn pod_collect_to_vec<
...
+  }
+
+  /// Convert a vec of the wrapper type into a vec of the inner type.
+  fn peel_vec(s: Vec<Self>) -> Vec<Inner>
+  where
+    Self: Sized,
+    Inner: Sized
+  {
+    // Not sure if these are needed
+    assert!(size_of::<*mut Inner>() == size_of::<*mut Self>());
```
notgull, line 334: Couldn't this be `assert_eq!`?

TheEdward162: I picked these up straight from here https://github.com/Lokathor/bytemuck/blob/e9eb13e0e75d00bab6c5227aa80e13ec82dd67be/src/transparent.rs#L162

Honestly I'm not sure why they are needed and was hoping it would become clearer in this PR. I _think_ the safety contract of `TransparentWrapper` should be enough for even a no-checks version.

Lokathor: Actually as someone else pointed out on discord, since Self and Inner are both Sized this will always pass.

but it will always get eliminated at compile time so maybe it's fine.

but also yes it could be assert_eq

src/allocation.rs
```diff
@@ -297,3 +297,55 @@ pub fn pod_collect_to_vec<
...
+    assert!(align_of::<*mut Inner>() == align_of::<*mut Self>());
+
+    let mut s = core::mem::ManuallyDrop::new(s);
+    unsafe {
+      // SAFETY:
+      // * ptr comes from Vec (and will not be double-dropped)
+      // * the two types have the identical representation
+      // * the len and capacity fields are valid
+      Vec::from_raw_parts(
+        s.as_mut_ptr() as *mut Self,
```
Lokathor, line 320: Someone reminded me that the call to .as_mut_ptr has to come *after* the calls to get the len and capacity, otherwise there's a stupid Stacked Borrows error and the pointer gets invalidated. I'd put each of these on their own line, and with as_mut_ptr as the last method to be called.

Lokathor:
this CI run will use the new and stricter miri flags but good luck to the PR

Lokathor (approved):

PR #117, opened 2022-07-04
Add `try_zeroed_vec` and `zeroed_vec`

fu5ha:
So, I believe this is sound as it essentially exactly mimics the internal implementation of `Vec::with_capacity` except using `alloc_zeroed` instead. However, `Vec::from_raw_parts_mut` is scary so would like to get someone else to review it and make sure I'm not going crazy.

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -11,7 +11,7 @@
 
 use super::*;
 use alloc::{
-  alloc::{alloc_zeroed, Layout, LayoutError},
+  alloc::{alloc_zeroed, Layout},
   boxed::Box,
   vec,
   vec::Vec,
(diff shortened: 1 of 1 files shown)


5225225:
I believe vec's allocation strategy is unspecified, so using from_raw_parts on something vec didn't allocate itself is questionable.

I'd do

```rust
let mut v = Vec::new();
v.try_reserve(length)?;
// somehow zero the vec, maybe make a zeroed item with T::zeroed() and write it to all items? or use `resize_with(T::zeroed)`
Ok(v)
```

src/allocation.rs
```diff
@@ -93,6 +93,31 @@ pub fn zeroed_box<T: Zeroable>() -> Box<T> {
   try_zeroed_box().unwrap()
 }
 
+/// Allocates a `Vec<T>` of length and capacity exactly equal to `length` and all elements zeroed.
+/// 
+/// ## Failure
+///
+/// This fails if the allocation fails, or if a layout cannot be calculated for the allocation.
+pub fn try_zeroed_vec<T: Zeroable>(length: usize) -> Result<Vec<T>, ()> {
```
notgull, line 101: I think this should be implemented using `try_zeroed_boxed_slice(length).into_vec()`. It would substantially reduce the unsafe code at play here.

src/allocation.rs
```diff
@@ -131,7 +151,7 @@ pub fn try_zeroed_slice_box<T: Zeroable>(
     let slice =
       unsafe { core::slice::from_raw_parts_mut(ptr as *mut T, length) };
```
notgull, line 152: This should be `core::ptr::slice_from_raw_parts_mut()` to avoid creating an intermediate reference (also also reduces the amount of unsafe by one line).

Nilstrieb:
> I believe vec's allocation strategy is unspecified
Sounds like a bug in vecs documentation. `Vec<T>` should always be allocated using the global allocator (as it currently works with the unstable allocator API), so in my opinion this should be fine.

Lokathor (approved):

PR #118, opened 2022-07-13
Replace LICENCES-* symlinks by a licence copy.

piwicode:
Fixes the licences content in bytemuck_derive.

Archives on crates.io does not contain a licence copy, but
just a file containing the symlink relative path instead.

https://static.crates.io/crates/bytemuck_derive/bytemuck_derive-1.1.0.crate

### DIFF
--- a/derive/LICENSE-APACHE
+++ b/derive/LICENSE-APACHE
@@ -1,1 +1,61 @@
-../LICENSE-APACHE
+Apache License
+Version 2.0, January 2004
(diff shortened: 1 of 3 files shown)


piwicode:
Thank you for the quick merge. 

Is there a chance that we can release a new version of bytemuck_derive in crates.io?
Let me know if I can help.

Lokathor:
1.1.1 released!

PR #119, opened 2022-07-22
Add `{wrap,peel}_box` functions to `TransparentWrapperAlloc`

WaffleLapkin:
This PR adds the following public API:
```rust
// TransparentWrapperAlloc<Inner: ?Sized>

fn wrap_box(s: Box<Inner>) -> Box<Self> { ... }
fn peel_box(s: Box<Self>) -> Box<Inner> { ... }
```

As a side-note: is it ok that `cargo fmt` changes formatting from the one in the repo?

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -339,6 +339,30 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>: TransparentWrapper<Inner> {
     }
   }
 
+  /// Convert a box to the inner type into a box to the wrapper
+  /// type.
+  #[inline]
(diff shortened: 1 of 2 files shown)


Lokathor:
yeah small format changes aren't a big deal.

Lokathor:
Can you throw some simple test cases of this being used in a tests file? Just so that the miri tests can hit the code a little.

WaffleLapkin:
I've added tests and also simplified tests that expect values (`{wrap,peel}_vec`), the casts there in the asserts seemed unnecessary, given that we are testing transparent wrapper there, not the casts.

Lokathor:
Released in 1.11

PR #124, opened 2022-08-07
Cleanup: use a macro for reprs, use syn errors & fix cfg-target-arch

danielhenrymantilla:
- Use a macro for the dozen options for `Repr`;
      - Use `impl Parse` on `Representation` for a simpler logic (no `Meta` "AST parsing" unnecessary complexity)
  - the previous bullet required unifying the `Error` types under `syn` (which paves the way for nicer span reporting in the future);
  - there was a `cfg(target_arch` on proc-macro code, and I'm unsure about the behavior there (it's a build lib after all), so I switch to emitting a `cfg_attr(target_arch…` instead just to be safe

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -5,8 +5,8 @@ extern crate proc_macro;
 mod traits;
 
 use proc_macro2::TokenStream;
-use quote::{quote, quote_spanned};
-use syn::{parse_macro_input, DeriveInput, spanned::Spanned};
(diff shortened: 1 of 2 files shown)


Lokathor:
@fu5ha you're the only other person i can recall having a slightly longer term interest in the proc-macros so i'm calling upon you to sanity check this.

fu5ha (approved):
Nice, lgtm

PR #125, opened 2022-08-11
Fix something_went_wrong compile error on SPIR-V

pali6:
The `Display` trait is conditionally removed from the error types on the SPIR-V architecture, yet `something_went_wrong` still has the trait bound. This results in the crate failing to compile for SPIR-V targets.

This PR conditionally disables the trait bound on the function on SPIR-V to fix this issue.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -20,22 +20,26 @@ possibility code branch.
 */
 
 /// Immediately panics.
+#[cfg(not(target_arch = "spirv"))]
 #[cold]
 #[inline(never)]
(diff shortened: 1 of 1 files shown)


Lokathor:
I'm immediately wondering why this is generic over Debug, don't we always call it with an error value from this very crate anyway? Can't we just make that the type it accepts?

But in the mean time yeah this seems good.

pali6:
According to the errors I got when compiling a shader it seems that it's only ever called with `CheckedCastError` and `PodCastError`. So there'd probably need to be some new trait for the two errors.

Issue #133, opened 2022-09-02
Unwieldy duplication of the public API


notgull:
As mentioned [here](https://github.com/Lokathor/bytemuck/pull/132#issuecomment-1234896922), `bytemuck`'s public API is very quickly becoming very unwieldy. Here's a screenshot of the current `docs.rs` main page. This doesn't include the functions in the `allocation` and `checked` modules.

![Screenshot_2022-09-01_21-01-13](https://user-images.githubusercontent.com/19805233/188056643-e5edc1cc-777a-47cc-a8af-f9c667127ff3.png)

My proposed solution is as follows:

Since `bytemuck` is essentially just a wrapper around "safe transmute", I propose that the current set of functions is replaced with a `try_cast` function and a `cast` function, that operates like this:

```rust
pub fn try_cast<Out, In: TransmutableTo<Out>>(input: In) -> Result<Out, In::Error> {
    if let Err(e) = input.can_transmute() {
        Err(e)
    } else {
        Ok(unsafe { transmute!(input) })
    }
}

pub fn cast<Out, In: TransmutableTo<Out>>(input: In) -> Out {
    try_cast(input).expect("Failed to transmute value")
}

pub unsafe trait TransmutableTo<T: Copy> : Copy {
    type Error;

    fn can_transmute(&self) -> Result<(), Self::Error>;
}
```

Where `try_cast` is basically just a wrapper around the `transmute!` macro.

The individual functions (i.e. `cast_ref` or `cast_slice`) could be replaced with `#[repr(transparent)]` helper structs. For instance, to emulate the current behavior of `cast`, there would be a `Raw` structure, implemented like so:

```rust
pub struct Raw<T>(pub T);

unsafe impl<In: NoUninit, Out: AnyBitPattern> TransmutableTo<Out> for Raw<In> {
    type Error = PodCastError;

    fn can_transmute(&self) -> Result<(), Self::Error> {
        // do the checks on size/alignment/et al
    }
}

// in user code
let a: u16 = bytemuck::cast(bytemuck::Raw(-3i16));
```

Then, you could implement `cast_ref` like so:

```rust
pub struct Ref<'a, T>(pub &'a T);

unsafe impl<'a, In: NoUninit, Out: AnyBitPattern> TransmutableTo<&'a Out> for Raw<'a In> {
    type Error = PodCastError;

    fn can_transmute(&self) -> Result<(), Self::Error> {
        // do the checks on size/alignment/et al
    }
}

[...]

PR #134, opened 2022-09-02
WIP/proof-of-concept: refactor `Box`/`Rc`/`Arc` to reduce code duplication.

zachs18:
WIP/proof-of-concept: As mentioned in #132, much of the code for `Box`/`Rc`/`Arc`'s functions is very similar. This WIP does not address the public API duplication mentioned in that PR and #133, but it does somewhat address the code duplication by factoring out the common code into generic `try_cast_aligned_ptr` and `try_cast_aligned_slice_ptr` functions, which use a trait `CastablePointer` to convert between smart pointer types and raw pointers.

The API is the same as #132 and is all implemented, but there are some "TODO"s in the code of things that could be added or things that could be removed.

Currently, the only changes are private, but something similar this could perhaps be made public.

Alternately to how it currrently works, the `CastablePointer<T>` trait could instead be parameterized on the "raw" type it returns, (rather than being hardcoded to have a "raw" type of `*mut T`). This could make it possible to additionally use it for `Vec`, though I don't know how useful that would be, since `Vec` needs some special handling anyway.

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -24,6 +24,67 @@ use core::{
   mem::ManuallyDrop,
   ops::{Deref, DerefMut},
 };
+use internal::{CastablePointer, OwnedCastablePointer};
+
+unsafe impl<T: ?Sized> CastablePointer<T> for Box<T> {
+  #[inline]
+  fn into_raw(self) -> *mut T {
+    Box::into_raw(self)
+  }
+
+  #[inline]
+  unsafe fn from_raw(ptr: *mut T) -> Self {
+    Box::from_raw(ptr)
+  }
+}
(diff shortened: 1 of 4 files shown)


zachs18:
Rebased, but I think I'm going to close this because:

1. `core::ptr::slice_from_raw_parts` was stabilized in 1.42.0, after bytemuck's MSRV, so `cast_{ref,mut,slice,slice_mut}` wouldn't even be able to be deduplicated with this.
2. while it does reduce code *duplication*, it also adds more code (most of which is comments, but still). Sometimes copy-pasting is more easier to understand than abstraction lol.

PR #137, opened 2022-10-23
Fix clippy lints

andrewgazelka:
- allows multiple clippy lints that fire from the derives in `bytemuck`

### DIFF
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -362,7 +362,11 @@ impl Derivable for Contiguous {
       quote!(),
       quote! {
           type Int = #integer_ty;
+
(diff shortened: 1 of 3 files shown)


Lokathor:
@andrewgazelka it looks like some of your changes were too aggressive because of the MSRV. For those parts, just silence the clippy lint if you like.

Lokathor:
Actually that lint I just allowed at the crate level, so just undo that change and get the updates from main and CI will probably pass.

Lokathor:
This is so old that the CI logs are gone it seems.

I'll accept it to allow the stuff then revert the `matches!` thing manually

PR #138, opened 2022-10-28
Fix `try_pod_read_unaligned` never having worked correctly

seritools:
It seems like a copy-paste error has happened between `try_from_bytes`/`try_from_bytes_mut` and `try_pod_read_unaligned`, causing `internal::try_pod_read_unaligned` to try to read a `&T::Bits` instead of a `T::Bits`, and thus usually failing with a `SizeMismatch` error.

In the worst case, this allows UB in safe code by having a type allowing any bit pattern and being pointer-sized.

### DIFF
--- a/derive/tests/basic.rs
+++ b/derive/tests/basic.rs
@@ -186,6 +186,21 @@ fn anybitpattern_implies_zeroable() {
   assert_eq!(test, AnyBitPatternTest { a: 0, b: 0 });
 }
 
+#[test]
+fn checkedbitpattern_try_pod_read_unaligned() {
+  let pod = [0u8];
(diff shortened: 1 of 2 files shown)


Lokathor (approved):

Lokathor:
This looks like a correct fix. I'm going to ask @fu5ha to double check since did the rework with the `internal` module stuff.

fu5ha (approved):
Nice find. Unfortunate that a seemingly small error like this would lead to bad behavior 🤔

Lokathor:
released as `1.12.2`

seritools:
Thanks for the quick merge and fix release!

PR #139, opened 2022-10-29
Allow `repr(transparent)` to be used generically in `derive(Pod)`

notgull:
My current use case is this:

```rust
#[derive(Pod, Zeroable)]
#[repr(transparent)]
struct MyNewtype<T>([T; 2]);
```

However, the current `derive` system only allows for `repr(packed)` to be used with generics, despite the fact that, by definition, `repr(transparent)` contains no padding. This PR adds the option for `repr(transparent)` to be used in this way.

I also moved to checks on the actual implementation to the `impl` block rather than being static, so that `MyNewtype` can also be used in the non-`Pod` case. I've verified that this works for my use case.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -9,7 +9,8 @@ use quote::quote;
 use syn::{parse_macro_input, DeriveInput, Result};
 
 use crate::traits::{
-  AnyBitPattern, Contiguous, Derivable, CheckedBitPattern, NoUninit, Pod, TransparentWrapper, Zeroable,
+  AnyBitPattern, CheckedBitPattern, Contiguous, Derivable, NoUninit, Pod,
+  TransparentWrapper, Zeroable,
 };
 
 /// Derive the `Pod` trait for a struct
@@ -56,8 +57,9 @@ pub fn derive_pod(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
 pub fn derive_anybitpattern(
   input: proc_macro::TokenStream,
 ) -> proc_macro::TokenStream {
-  let expanded =
-    derive_marker_trait::<AnyBitPattern>(parse_macro_input!(input as DeriveInput));
+  let expanded = derive_marker_trait::<AnyBitPattern>(parse_macro_input!(
+    input as DeriveInput
+  ));
 
   proc_macro::TokenStream::from(expanded)
 }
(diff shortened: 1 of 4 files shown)


derive/src/lib.rs
```diff
@@ -252,10 +254,57 @@ fn derive_marker_trait_inner<Trait: Derivable>(
...
+  trait_name: &syn::Path
+) {
+  // Get each generic type parameter.
+  let type_params =generics.type_params().map(|param| &param.ident).cloned().collect::<Vec<_>>();
+
+  // Add a where clause if it doesn't already exist.
+  let where_clause = generics.where_clause.get_or_insert_with(|| syn::WhereClause {
+    where_token: Default::default(),
+    predicates: Default::default(),
+  });
```
danielhenrymantilla, line 277:
```suggestion
  let where_clause = generics.make_where_clause();
```

derive/src/lib.rs
```diff
@@ -252,10 +254,57 @@ fn derive_marker_trait_inner<Trait: Derivable>(
...
+  );
+}
+
+fn path_from_ident(ident: &syn::Ident) -> syn::Path {
+  syn::Path {
+    leading_colon: None,
+    segments: vec![syn::PathSegment {
+      ident: ident.clone(),
+      arguments: syn::PathArguments::None,
+    }]
```
danielhenrymantilla, line 306:
```suggestion
    segments: Some(syn::PathSegment {
      ident: ident.clone(),
      arguments: syn::PathArguments::None,
    })
```

derive/src/lib.rs
```diff
@@ -252,10 +254,57 @@ fn derive_marker_trait_inner<Trait: Derivable>(
...
 
     #implies_trait
   })
 }
+
+/// Add a trait marker to the generics if it is not already present
+fn add_trait_marker(
+  generics: &mut syn::Generics,
+  trait_name: &syn::Path
+) {
```
danielhenrymantilla, line 269:
```rs
fn add_trait_marker(
  generics: &mut syn::Generics,
  trait_name: &syn::Path
) {
    let type_params: Vec<WherePredicate> =
        generics.type_params().map(|p| &p.ident).map(|ty| parse_quote!(
            #ty : #trait_name
        )).collect();
    generics.make_where_clause().predicates.extend(type_params);
}
```

These 5 lines seem more straightforward to read/maintain than your 40 lines heroically bypassing `parse_quote!` (I find the maintainability benefits of `parse_quote!` to outweigh the sub-milli-second runtime-of-the-macro overhead of parsing) 🙃

danielhenrymantilla (approved):
LGTM ✅

fu5ha (approved):
The idea seems good to me, will defer to @danielhenrymantilla on implementation review :)

derive/src/lib.rs
notgull: That's definitely a better approach than mine. I've implemented it.

danielhenrymantilla (approved):
The new `add_trait_marker` LGTM ✅; just note that there are a bunch of `rustfmt`-induced changes as well in that last commit

notgull:
Whoops, I guess I just run rustfmt before commit by instinct nowadays.

Lokathor:
rustfmt is no big deal, i do the same thing.

Lokathor:
Released `bytemuck_derive-1.3.0`

PR #140, opened 2022-11-04
Conditionally compile in `Arc`

CryZe:
Apparently the latest release of `bytemuck` doesn't compile on some of the targets that it used to because it uses `Arc`. `Arc` is not a type that exists on every target, because not every target supports atomics.

Error message with 1.12.2:
```
error[E0432]: unresolved import `alloc::sync`
  --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/bytemuck-1.12.2/src/allocation.rs:17:3
   |
17 |   sync::Arc,
   |   ^^^^ could not find `sync` in `alloc`
```

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -10,11 +10,12 @@
 //! ["extern_crate_alloc"]}`
 
 use super::*;
+#[cfg(target_has_atomic = "ptr")]
+use alloc::sync::Arc;
 use alloc::{
   alloc::{alloc_zeroed, Layout},
   boxed::Box,
   rc::Rc,
-  sync::Arc,
   vec,
   vec::Vec,
 };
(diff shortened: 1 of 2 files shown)


notgull (commented):
Correct me if I'm wrong, but wasn't `target_has_atomic` introduced in a Rust version after 1.34? This would break the MSRV.

CryZe:
The `alloc` feature does not follow the MSRV.

Lokathor:
That's accurate, the CI isn't that precise with the MSRV. Only the "basic" build sticks exactly to 1.34. Everything else is only tested on latest stable.

I'd rather not create *even more* features for people to have to manage, if it's reasonable to just keep two features together. In the case of `target_has_atomic`, it seems to be the "intended' cfg to check for this sort of thing. According to <https://github.com/rust-lang/rust/issues/32976> it's been stable since 1.60 back in Feb, which seems "long enough" that we can generally assume people are using at least that version. That's new enough that it's not in Debian Stable, but it is in Debian Testing, which should become the stable version sometime in 2023.

I'm inclined to accept this as is, unless @notgull you really think we should break it up?

notgull:
I'm fine with this!

Lokathor:
I agree, but I'm not sure I have the energy for that much CI fiddling

CryZe:
Can we have a bugfix release for this? My CI blocks me otherwise xD

Lokathor:
Yeah, I had enough time to hit the button before the bus to work but can't sit down at a publish capable thing until some time later today.

CryZe:
Ah no worries then :D

Lokathor:
1.12.3 should be available now

CryZe:
Nice, everything works now 👍 
Thank you :)

PR #141, opened 2022-11-04
Move the array impl for `Pod` to superset types

GabrielDertoni:
This change removes the current `impl Pod for [T; N]` and moves it to superset types, namely into two other impls: `impl NonUninit for [T; N]` and `impl AnyBitPattern for [T; N]`.

This, makes it possible for the user to cast between an array of `NonUnint` and an array of `AnyBitPattern` for example, which was previously impossible. One example usecase would be to cast from`[char; N]` to `[u32; N]`.

I think this shouldn't break any existing code, since before only code with `T: Pod` would compile, and now both `T: NonUninit` and `T: AnyBitPattern` works, and those are strict supersets of `Pod`.

### DIFF
--- a/src/anybitpattern.rs
+++ b/src/anybitpattern.rs
@@ -54,3 +54,13 @@ pub unsafe trait AnyBitPattern:
 }
 
 unsafe impl<T: Pod> AnyBitPattern for T {}
+
+#[cfg(not(feature = "min_const_generics"))]
+impl_unsafe_marker_for_array!(
+  AnyBitPattern, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19,
+  20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 48, 64, 96, 128, 256,
+  512, 1024, 2048, 4096
+);
+
+#[cfg(feature = "min_const_generics")]
+unsafe impl<T, const N: usize> AnyBitPattern for [T; N] where T: AnyBitPattern {}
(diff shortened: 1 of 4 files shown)


notgull:
If I had a function:

```rust
fn foobar<T: Pod>(t: T) { /* ... */ }
```

and I used it like this:

```rust
foobar([1, 2, 3]);
```

it would fail to compile, since `[i32; 3]` no longer implements `Pod`.

Lokathor:
Yeah we certainly can't remove the impls for Pod in version 1 of the crate. It's unfortunate, but that type was defined first and so plenty of code *only* uses the Pod trait, not the more specific traits. Hopefully this can be fixed in version 2 of the crate (which won't be fully designed until the safe-transmute project stabilizes and we find out what exactly it will allow).

GabrielDertoni:
Oh, I see. Well, I guess ideally `Pod` should be defined as

```rust
pub unsafe trait Pod: NoUninit + AnyBitPattern + Copy + 'static { }
```

And then the impls `impl<T: Pod> NoUninit for T { }` and `impl<T: Pod> AnyBitPattern for T { }` would be removed. Allowing the impls `impl<T: Pod> Pod for [T; N]`, `impl<T: NoUninit> NoUninit for [T; N]` and `impl<T: AnyBitPattern> AnyBitPattern for [T; N]` to coexist.

However the problem now, of course, is that it is a huge breaking change where anyone using bytemuck for custom types would have to then use it like

```rust
#[derive(NoUninit, AnyBitPattern, Pod)]
struct Foo { /* ... */ }
```

I personally think that this is ok since it is similar to something like `PartialOrd` with `Ord` and `PartialEq` with `Eq`.

Lokathor:
Since this would be a big breaking change, i'll just close it for now. In general, i'd rather not do any breaking change within the crate before the safe-transmute RFC stabilizes.

PR #144, opened 2022-11-09
Implement `ByteEq` and `ByteHash` derives

### LINKED ISSUE #81: Derives for `Eq` and `Hash`
It probably makes sense to provide derives for `Eq` (and `PartialEq`) and `Hash` in bytemuck. On a `Pod` type `Eq` can be a simple memcmp / bcmp. Hashing is also much more efficient as you can provide all bytes at once.

CryZe:
This adds the derives `ByteEq` and `ByteHash` that can be used as an alternative to the `Eq` / `PartialEq` and `Hash` derives from the standard library. The difference is that these variants use `bytemuck` to convert their values to byte slices before comparing / hashing them. This allows the comparisons to turn into a simple `memcmp` / `bcmp` (or completely inlined as a few vector instructions) and allows hashers to process all bytes at once, possibly allowing for some vector operations as well.

Here's a quick comparison of the generated assembly: 

Resolves #81

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -224,6 +224,87 @@ pub fn derive_contiguous(
   proc_macro::TokenStream::from(expanded)
 }
 
+/// Derive the `PartialEq` and `Eq` trait for a type
+///
+/// The macro implements `PartialEq` and `Eq` by casting both sides of the
+/// comparison to a byte slice and then compares those.
+///
+/// ## Warning
+///
+/// Since this implements a byte wise comparison, the behavior of floating point
+/// numbers does not match their usual comparison behavior. Additionally other
+/// custom comparison behaviors of the individual fields are also ignored.
+///
+/// ## Example
+///
+/// ```rust
+/// # use bytemuck_derive::{ByteEq, NoUninit};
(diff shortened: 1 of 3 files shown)


notgull:
I'm a fan of this idea. Truth be told I though that the code in the second image would optimize to the first one, but apparently not.

Lokathor:
I would add another paragraph to the warnings on ByteEq that using this derive instead of the `core` derive does *not* give you a [StructuralEq](https://doc.rust-lang.org/nightly/core/marker/trait.StructuralEq.html) impl.

and probably we want `inline` and `must_use` on the `eq` method, as well as `inline` on the hash methods.

CryZe:
Alright I added those suggestions.

CryZe:
Is this blocked on anything?

Lokathor (approved):

CryZe:
Thanks :D

Lokathor:
Sorry, haven't had a time for Rust in a bit and it fell behind.

I'll try to do a release later

CryZe:
I don't need this "urgently", so feel free to also just not do a release :)

PR #146, opened 2022-11-16
Remove where clause from derived TransparentWrapper impls.

### LINKED ISSUE #145: `TransparentWrapper` derive emits unnecessary bound on generic structs.
```rs
#[derive(bytemuck::TransparentWrapper)]
#[repr(transparent)]
pub struct GenericWrapper<T>(pub T);
```
This emits
```rs
unsafe impl<T> ::bytemuck::TransparentWrapper<T> for GenericWrapper<T> where
    T: ::bytemuck::TransparentWrapper<T>
{
}
```
The `where T: ::bytemuck::TransparentWrapper<T>` is not necessary, and is not implemented by default (there is no blanket `impl<T> TransparentWrapper<T> for T`), so it makes `derive(TransparentWrapper)` on generic structs not very useful.

---

Adding a blanket `impl<T> TransparentWrapper<T> for T` or removing the `where T: bytemuck::TransparentWrapper<T>` bound from the derived impl would solve the issue.

I think adding blanket would be a breaking change, but removing the bound on the derived impl should not be.

zachs18:
Remove the unnecessary `where T: TransparentWrapper<T>` bound on derived TransparentWrapper impls for generic structs.
Resolves #145

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -252,6 +252,12 @@ fn derive_marker_trait_inner<Trait: Derivable>(
     quote!()
   };
 
+  let where_clause = if Trait::requires_where_clause() {
+    where_clause
+  } else {
+    None
+  };
+
   Ok(quote! {
     #asserts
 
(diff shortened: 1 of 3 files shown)


Lokathor:
This seems reasonable. Can you add a test so that we don't regress this again?

zachs18:
Sure, I added a test using just a single struct member, and one using `#[transparent(T)]`.

Lokathor (approved):

PR #147, opened 2022-11-16
Extend `#[derive(TransparentWrapper)]`

zachs18:
(Based on PR 146, but not particularly related)
Allow types more complicated than single idents in `#[transparent(Type)]` when deriving `TransparentWrapper`.
   * It still does string comparison, because resolving type equality is not possible in a proc-macro.
   * This is valid as long as there is no way to make the same token stream/source string refer to different types in the struct definition and the derived impl, which I believe is the case.
     * Maybe some unnamed const or function could be emitted to ensure the two types are actually the same at compile-time? like `const _: () = { fn _detail<Generics>(x: (Type1, Type2), _generic_use: PhantomData<Struct<Generics>>) -> (Type2, Type1) { x } };`

TODO: think about and verify that variance cannot break things.

### DIFF
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -487,7 +487,8 @@ impl Derivable for TransparentWrapper {
           if explicit {
             bail!("TransparentWrapper must have one field of the wrapped type. \
                    The type given in `#[transparent(Type)]` must match tokenwise \
-                   with the type in the struct definition, not just be the same type");
+                   with the type in the struct definition, not just be the same type. \
+                   You may be able to use a type alias to work around this limitation.");
           } else {
             bail!("TransparentWrapper must have one field of the wrapped type");
           }


zachs18:
Since users can in most cases just do like

```Rust
type Wrapped = Complex<Type>;
#[derive(TransparentWrapper)]
#[transparent(Wrapped)]
#[repr(transparent)]
struct MyType {
  wrapped: Wrapped,
  etc: ZST,
}
```

I'm not sure if this is worth the trouble. Except I guess complex types involving generics are not representable without this, e.g.

```Rust
#[derive(TransparentWrapper)]
#[transparent(Wrapped)]
#[repr([T; 2])]
struct ArrayPair<T> {
  wrapped: [T; 2],
  etc: ZST,
}
```

Lokathor:
`#[repr([T; 2])]`?

zachs18:
Typo

#[derive(TransparentWrapper)]
#[transparent([T; 2])]
#[repr(transparent)]
struct ArrayPair<T> {
  wrapped: [T; 2],
  etc: ZST,
}

derive/src/traits.rs
```diff
@@ -534,24 +557,31 @@ fn generate_fields_are_trait(
...
+  let mut tokens = tokens.into_iter().peekable();
+  match tokens.peek() {
+    Some(TokenTree::Group(group)) => {
+      let res = get_wrapped_type_from_stream(group.stream());
+      tokens.next();
+      match tokens.next() {
+        // If there were more tokens, the input was invalid
+        Some(_) => None,
+        None => res,
+    }
```
fu5ha, line 570: weird formatting

derive/src/traits.rs
```diff
@@ -271,15 +282,23 @@ impl Derivable for TransparentWrapper {
 
   fn asserts(input: &DeriveInput) -> Result<TokenStream> {
     let fields = get_struct_fields(input)?;
-    let wrapped_type = match Self::get_wrapper_type(&input.attrs, &fields) {
-      Some(wrapped_type) => wrapped_type.to_string(),
+    let (wrapped_type, explicit) = match Self::get_wrapped_type(&input.attrs, &fields) {
+      Some(WrappedType { wrapped_type, explicit }) => (wrapped_type.to_token_stream().to_string(), explicit),
       None => unreachable!(), /* other code will already reject this derive */
     };
+    dbg!(&wrapped_type);
```
fu5ha, line 289: Extra dbg

derive/src/traits.rs
```diff
@@ -271,15 +282,23 @@ impl Derivable for TransparentWrapper {
...
     let mut wrapped_fields = fields
       .iter()
-      .filter(|field| field.ty.to_token_stream().to_string() == wrapped_type);
+      .filter(|field| dbg!(field.ty.to_token_stream().to_string()) == wrapped_type);
```
fu5ha, line 292: Extra dbg

derive/src/traits.rs
```diff
@@ -271,15 +282,23 @@ impl Derivable for TransparentWrapper {
...
     if let None = wrapped_fields.next() {
-      bail!("TransparentWrapper must have one field of the wrapped type");
+      if explicit {
+        bail!("TransparentWrapper must have one field of the wrapped type. \
+               The type given in `#[transparent(Type)]` must match tokenwise\
+               with the type in the struct definition, not just be the same type");
```
fu5ha, line 297: We could suggest the type alias workaround here as well.

fu5ha (changes requested):

derive/src/traits.rs
zachs18: Formatting fixed.

For the semantics of this section though, we might want to allow a trailing comma here. As-is, this rejects `#[transparent(T, )]`, which might be useful if `T` is long and would naturally spill onto another line.

Also, if the parsing of the `transparent` helper fails, the code falls back to the "find the one field" strategy, e.g. this succeeds:

```rs
#[derive(bytemuck::TransparentWrapper)]
#[repr(transparent)]
#[transparent(not a type)]
struct Foo {
 one_field: u32,
};
```

which is not ideal. Maybe `get_type_from_simple_attr` should bail if it finds the attribute but fails to parse it?

Lokathor (approved):

PR #148, opened 2022-11-19
Implement `Zeroable` on `UnsafeCell<impl Zeroable>`

yvt:
`UnsafeCell<T>` is `repr(transparent)`, and its constructor is not `unsafe`, hence it should be safe to implement `Zeroable` on it, provided that `T` is also `Zeroable`.


notgull:
This looks good to me. While `Pod` on `UnsafeCell` is a non-starter, I think it should always be safe to zero it out if the interior can be too.

Lokathor (approved):

notgull:
I think we should do this for the other transparent std types as well, like `Cell` or `Reverse`.

notgull:
Was about to do this but it looks like I got sniped by #149

PR #149, opened 2022-11-21
Implement `Zeroable` on `{Cell,Reverse}<impl Zeroable>`

yvt:
Implements `Zeroable` on additional types provided by Rust 1.34.


Lokathor (approved):

notgull:
Can't `Reverse` also implement `Pod`?

Lokathor:
It could, same as Wrapping does

PR #151, opened 2022-11-29
Implement CheckedBitPattern for core::num::NonZero*.

zachs18:
Also adds tests for `CheckedBitPattern`.

Also fixes a possible Stacked-Borrows violation in `checked::try_cast_slice_mut` found via miri (this can be split off if necessary).

TODO: 
* [ ] Determine if this is guaranteed to be sound.

---

This is sound based on the current implementation of `num::NonZero[int]`, since they're `repr(transparent)` wrappers around `[int]` (with some `rustc_` magic saying that `0` is an invalid bit-pattern).

I'm not sure if Rust documentation explicitly guarantees that `NonZero*` has the same layout as the corresponding primitive integer type, but IMO the intent is that they do, since ~~anything else would be dubious~~ it does say:
* "For example, `Option<NonZero[int]>` is the same size as `[int]`" (in the docs for each `num::NonZero[int]`)
* and "Rust guarantees to optimize the following types `T` such that `Option<T>` has the same size as `T`: [...] `num::NonZero*`" (in the docs for std::option).

By the above two, we know that `NonZero[int]` and `[int]` have the same size. Nothing is explicitly stated about alignment or representation that I've found.

### DIFF
--- a/src/checked.rs
+++ b/src/checked.rs
@@ -395,7 +395,7 @@ pub fn try_cast_slice_mut<
 
   if pod.iter().all(|pod| <B as CheckedBitPattern>::is_valid_bit_pattern(pod)) {
     Ok(unsafe {
-      core::slice::from_raw_parts_mut(pod.as_ptr() as *mut B, pod.len())
+      core::slice::from_raw_parts_mut(pod.as_mut_ptr() as *mut B, pod.len())
     })
   } else {
     Err(CheckedCastError::InvalidBitPattern)
--- a/src/checked.rs
+++ b/src/checked.rs
@@ -170,6 +170,35 @@ unsafe impl CheckedBitPattern for bool {
   }
 }
 
+macro_rules! impl_checked_for_nonzero {
+  ($($nonzero:ty: $primitive:ty),* $(,)?) => {
+    $(
+      unsafe impl CheckedBitPattern for $nonzero {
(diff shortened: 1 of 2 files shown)


Lokathor:
This looks correct to me, but I'm also gonna ping @fu5ha

In terms of the soundness, instead of just checking for non-zero, it can also check that size_of and align_of are equal for the base type and the NonZero type (eg: size_match && align_match && not_zero). That'll just const fold away on all current targets, while still catching the problem if there is a mismatch on some future setup.

fu5ha (approved):
Lgtm :)

Lokathor (approved):

zachs18:
> In terms of the soundness, instead of just checking for non-zero, it can also check that size_of and align_of are equal for the base type and the NonZero type.

Sounds good to me. I also added a test that will fail if that is ever false for any `NonZero*` (on the testing platform).

Lokathor:
I'll try to release within the next day.

Lokathor:
1.12.4 should be out now

Lokathor:
actually i guess there was a problem with the derives also needing a release so i had to yank it, issue here https://github.com/Lokathor/bytemuck/issues/165

yanchith:
Ugh, sorry to cause trouble, and thank you ❤️

PR #152, opened 2022-12-03
Implement `AnyBitPattern` for `MaybeUninit<T>`

### LINKED ISSUE #108: Could `MaybeUninit<T: Copy>` implement `AnyBitPattern`?
It already implements `Zeroable`, and the main draw of `MaybeUninit` is that it can contain theoretically any bit pattern.

zachs18:
... `where MaybeUninit<T>: Copy + 'static`.

Resolves #108 .

A feature flag is required because `MaybeUninit` was unstable in bytemuck's MSRV. I just used the `#[cfg(feature = "zeroable_maybe_uninit")]` flag since that's required anyway, and it didn't seem reasonable to add another feature flag for this.

The `where MaybeUninit<T>: Copy + 'static` bound is required because of `AnyBitPattern`'s supertrait bounds `Copy + 'static`. A simpler bound would be `where T: Copy + 'static` (i.e. have the bound on `T` instead of `MaybeUninit<T>`). These are currently (and probably always will be) equivalent, but if `std` ever relaxes the `Copy` impl of `MaybeUninit`, this bound will reflect that, whereas `where T: Copy + 'static` would not.

### DIFF
--- a/src/anybitpattern.rs
+++ b/src/anybitpattern.rs
@@ -54,3 +54,9 @@ pub unsafe trait AnyBitPattern:
 }
 
 unsafe impl<T: Pod> AnyBitPattern for T {}
+
+#[cfg(feature = "zeroable_maybe_uninit")]
+unsafe impl<T> AnyBitPattern for core::mem::MaybeUninit<T> where
+  core::mem::MaybeUninit<T>: Copy + 'static
+{
+}


zachs18:
Note about `MaybeUninit<InteriorMutableType>`:  `AnyBitPattern` requires no interior mutability, but there's no safe way to access the interior mutability with just `MaybeUninit`, so I'm inclined to say it's fine. The docs for `AnyBitPattern` say
> More precisely: A shared reference to the type must allow reads, and only reads.

This is true of `&MaybeUninit<T>` when only considering its *safe* API. `fn as_mut_ptr` gives you a `*mut T`, which is unsafe to write through.

zachs18:
So it turns out that `Cell<T>` isn't `Copy` probably not for soundness reasons, but just as a lint. So this PR is probably unsound as written. Closing for now.

https://github.com/rust-lang/rust/pull/55207

TODO: Maybe `where MaybeUninit<T>: Sync + Copy + 'static` would be valid, for the reasoning above? Alternately this could be added under a new `unsound_maybeuninit_anybitpattern_impl` as per `unsound_ptr_pod_impl`, though IIUC that was only added as a semver exception.

zachs18:
This might still be sound if `AnyBitPattern`'s requirements for interior mutability are clarified to only apply to safe code (since accessing the `T` in a `MaybeUninit<T>` requires unsafe code), otherwise this is unsound I think.

Lokathor:
Pod itself also assumes Freeze without saying so.

Really, Freeze needs to be a trait that code can talk about directly

PR #157, opened 2022-12-19
`impl Zeroable for Atomic*`

### LINKED ISSUE #74: Zeroable should be implemented for `Atomic*`
The atomic types are all bytemuck-compatible, and in fact as far as I know all bytemuck traits can be implemented for `Atomic*` that can be implemented for the underlying integer type as the operations available on the atomic types are a strict superset (i.e. setting their value via an immutable reference). Even casting between a non-atomic int and an atomic int should be safe, although from this discussion here https://github.com/rust-lang/rust/issues/76314 it may be invalid to cast an `&usize` (or other scalar) to an `&AtomicUsize` (or other atomic scalar). This sadly means that `Pod` cannot be implemented without adding new traits to account for types where owned types can be cast but references cannot.

nolanderc:
fixes #74

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ derive = ["bytemuck_derive"]
 extern_crate_alloc = []
 extern_crate_std = ["extern_crate_alloc"]
 zeroable_maybe_uninit = []
+zeroable_atomics = []
 min_const_generics = []
 wasm_simd = [] # Until >= 1.54.0 is MSRV this is an off-by-default feature.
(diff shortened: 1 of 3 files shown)


nolanderc:
I see!

Pushed a new commit with the feature flag. Documented the feature (and the other feature flags in that module) on the trait. 

Also, I saw that there was also some bad references in `checked.rs`, so I fixed these as well.

Lokathor (approved):

PR #160, opened 2023-01-05
Implement AnyBitPattern for MaybeUninit<T> where T: AnyBitPattern.

### LINKED ISSUE #108: Could `MaybeUninit<T: Copy>` implement `AnyBitPattern`?

zachs18:
(Partially) resolves #108 .

A feature flag is required because `MaybeUninit` was unstable in bytemuck's MSRV. I just used the `#[cfg(feature = "zeroable_maybe_uninit")]` flag since that's required anyway, and it didn't seem reasonable to add another feature flag for this.

---

As discussed in #152 , it is probably unsound to have `MaybeUninit<T>` implement `AnyBitPattern` when `T` contains interior mutability, so a blanket `impl<T: Copy + 'static> AnyBitPattern for MaybeUninit<T>` would be unsound under that reasoning. However, if `T` implements `AnyBitPattern`, then it is known to not have any interior mutability, so `T: AnyBitPattern` is a sufficient (but not necessary) condition for `MaybeUninit<T>: AnyBitPattern` to be sound.

Since `T: AnyBitPattern`  sufficient-but-not-necessary requirement, it could be relaxed later, e.g. if APIs around `Freeze` become available.

### DIFF
--- a/src/anybitpattern.rs
+++ b/src/anybitpattern.rs
@@ -54,3 +54,7 @@ pub unsafe trait AnyBitPattern:
 }
 
 unsafe impl<T: Pod> AnyBitPattern for T {}
+
+#[cfg(feature = "zeroable_maybe_uninit")]
+unsafe impl<T> AnyBitPattern for core::mem::MaybeUninit<T> where T: AnyBitPattern
+{}


Lokathor (approved):

Lokathor:
Follow up question: Can't you cast *from* AnyBitPattern *to* AnyBitPattern? Like that's the idea right? Then, wouldn't this impl allow you to cast from MaybeUninit *to* something else? Because that could put uninit bytes into the something else where they're not supposed to be.

zachs18:
No, I think you can only cast from NoUninit to AnyBitPattern. AnyBitPattern does not alone imply NoUninit.

> The requirements for this is very similar to Pod, except that the type can allow uninit (or padding) bytes.

(from [docs](https://docs.rs/bytemuck/1.12.2/bytemuck/trait.AnyBitPattern.html))

Lokathor:
https://github.com/Lokathor/bytemuck/pull/150 hm, but couldn't you use `pod_read_unaligned`?

zachs18:
No, `pod_read_unaligned` (and the `try_`/`checked::` versions) take in a `&[u8]`, which cannot be uninitialized.

Lokathor:
uhhh, yeah you're right, you're right. Sorry for the extra questions, MaybeUninit just makes me extra paranoid.

PR #176, opened 2023-02-26
Use `align_offset` to check alignment

sarah-ek:
check for alignment using `(ptr as *const ()).align_offset(align) == 0`
This is in a way better than `ptr as usize % align == 0`, because casting a pointer to an integer has the side effect that it exposes the pointer's provenance, which may theoretically inhibit some compiler optimizations.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -130,6 +130,19 @@ pub(crate) unsafe fn pod_read_unaligned<T: Copy>(bytes: &[u8]) -> T {
   }
 }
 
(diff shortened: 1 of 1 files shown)


Lokathor:
Since this method isn't available in 1.34 we'd have to put the change behind a feature gate.

But also, a compiler "should" be able to see that even the integer form of the pointer doesn't escape and so I would hope that the cast does not hurt optimizations in practice.

Lokathor:
I want the basics of the PR, but since sarah hasn't replied i'll just accept it as-is and fix it up with a new feature.

Issue #177, opened 2023-02-28
Should `pod_collect_to_vec()` input type require `AnyBitPattern`?


MarijnS95:
Disclaimer: I'm still fuzzy on the various traits and their requirements.

`pod_collect_to_vec()` requires `NoUninit + AnyBitPattern` on both the source slice and output `Vec` type. However, casts like `cast_vec` and `cast_slice` only require `NoUninit` on the source type (we can safely read all bytes, but don't need to zero-initialize the source type) and require the full `AnyBitPattern` exclusively on the output type (for `pod_collect_to_vec()`: to first zero-initialize the output `Vec` before copying into it?).

https://github.com/Lokathor/bytemuck/blob/1039388f0baa4532f499c7e809a92dc5552e2edd/src/allocation.rs#L285-L288

Is this constraint desired (and if so, can it be more clearly documented?) or accidental (and can it be relaxed)?

Thanks for this crate by the way! It's solving a bunch of problems for us that have always been looming around the corner :grimacing:

Lokathor:
The reason for the `NoUninit` requirement on the destination type is because of the guts of the function itself:

```rust
pub fn pod_collect_to_vec<
  A: NoUninit + AnyBitPattern,
  B: NoUninit + AnyBitPattern,
>(
  src: &[A],
) -> Vec<B> {
  let src_size = size_of_val(src);
  // Note(Lokathor): dst_count is rounded up so that the dest will always be at
  // least as many bytes as the src.
  let dst_count = src_size / size_of::<B>()
    + if src_size % size_of::<B>() != 0 { 1 } else { 0 };
  let mut dst = vec![B::zeroed(); dst_count];

  let src_bytes: &[u8] = cast_slice(src);
  let dst_bytes: &mut [u8] = cast_slice_mut(&mut dst[..]);
  dst_bytes[..src_size].copy_from_slice(src_bytes);
  dst
}
```

The actual copy from the `Vec<A>` into the new `Vec<B>` happens on two `[u8]` references that are equal sized. However, we can't make both those `[u8]` *references* if either type has uninit bytes.

If we really had to, we could use slice *pointers* instead of slice references, but then the function itself would have to do one or two unsafe calls. Even within `bytemuck`, it's *preferred* when a function doesn't need to open up any unsafe of its own and we can just rely on the other traits/functions.

[...]

PR #178, opened 2023-02-28
Fix 177

Lokathor:
Closes https://github.com/Lokathor/bytemuck/issues/177

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -209,7 +209,7 @@ pub fn cast_vec<A: NoUninit, B: AnyBitPattern>(input: Vec<A>) -> Vec<B> {
 ///   alignment.
 /// * The start and end content size in bytes of the `Vec` must be the exact
 ///   same.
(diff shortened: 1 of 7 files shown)


MarijnS95 (approved):
Yup thanks, that is exactly what I was looking for!

I thought I saw the same constraint on other functions, but would have to recheck if/where, and if they may be intended.

Lokathor:
If you find more cases feel free to open more issues, but i'll merge and publish this update for now.

MarijnS95:
That's most appreciated, while I'm migrating to `bytemuck` :)

Lokathor:
released `bytemuck-1.13.1`. As I write this it's on crates.io, but it'll be ages before docs.rs shows the new docs (252 in queue, so like a 12 hour wait just if every run takes only 3 minutes each, docs.rs has been clogged as heck lately)

MarijnS95:
Yup, thanks! Already using it, and already don't need it after switching to `cast_slice().to_vec()` as discussed to likely be simpler and more performant :)

PR #183, opened 2023-03-21
Switch to `syn` 2.0

CryZe:
This upgrades the derives from `syn` 1.0 to `syn` 2.0. While the MSRV for `syn` 2.0 is not Rust 1.56 which is higher than the MSRV of `bytemuck`, the derives don't fall under the MSRV policy.


Lokathor (approved):
What an almost entirely painless major version upgrade.

PR #184, opened 2023-03-25
Add generic way of converting the item type of a container

### LINKED ISSUE #133: Unwieldy duplication of the public API

### LINKED ISSUE #136: convert 2d array to slice without trait bounds
right now you can convert a `&[[T; N]; M]` into `&[T]` *when T is Pod*, but we should add a function for this even for any generic `T`.

`array2_as_slice` or something, I dunno

### LINKED ISSUE #167: Getting allocated bytes (Box/Rc/Arc) out of Box<T>
When working with allocated types, I often wish for a function like [`bytemuck::bytes_of`](https://docs.rs/bytemuck/latest/bytemuck/fn.bytes_of.html) but for heap types.

For example, if I have a compatible `Box<Metadata>`, I should be able to receive a zero-cost `Box<[u8]>` and so on.

For now, the only casts available are either [between sized Boxes](https://docs.rs/bytemuck/latest/bytemuck/allocation/fn.cast_box.html) or between [`Vec`s of different types](https://docs.rs/bytemuck/latest/bytemuck/allocation/fn.cast_vec.html) but I couldn't find a way to go from `Box<T>` to `Vec<u8>` or anything similar.

Jarcho:
This is capable of replacing all the `cast_*` functions except the `cast` itself. No more need for `cast_ref`, `cast_slice` `cast_box`, `cast_vec` and such and any mutable variants. Also moves some errors to post-mono compiler errors rather than runtime errors.

This is not implementable with GAT's, so this has an msrv of `1.65`.

Docs and testing still need to be done, as well as implementing all the containers in `alloc`.

fixes #133
fixes #167
fixes #136


Lokathor:
CI failed on the 1.34 version. This whole deal would have to go behind a feature gate, given the crate's MSRV.

That said, I'm extremely intrigued by this proposal. I was initially skeptical how you'd have it handle the difference between owned containers and borrowed containers, but that seems to be captured by the pair of "CastPtr" and "CastMetadata". I'm going to link this in the `#dark-arts` channel of the community discord and perhaps other places to get a lot of eyes on it, a whole new 500 line unsafe mini-framework is a lot to go over for just one person.

Jarcho:
Ideally `CastMetadata` would only need to exist for `Vec<_>` and `CastPtr` would handle slice pointers, but there's currently no safe way to get the length from them. This would remove all the duplication the comes from having to treat `Container<_>` and `Container<[_]>` differently.

There difference between owned and borrowed containers is handled by the `Ctor` and `Container` traits. They essentially abstracts away unwrapping a container to it's raw parts, and re-wrapping it with a different item type.

`Ctor` essentially emulates HKT's by providing a different implementation for each container. This lets us refer to just the container (e.g. `Box`), rather than a container with a specific item type (e.g. `Box<u32>`).

zachs18:
Wow! Great concept here! My old draft PR #134 was similar but only handled private API and didn't go this in-depth, but could still perhaps be useful to look at.

---

~~I got [nerd-sniped](https://xkcd.com/356/) a bit, so~~ I added `Option<C>`, `Box`/`Arc`/`Rc<T>`/`<[T]>`, and `Vec<T>` support on [my fork](https://github.com/zachs18/bytemuck/tree/abs_container) and added some tests (Feel free to merge/cherry-pick. See below for why I added the `*Ptr` types to implement `CastPtr`).

Currently the `PodCastError` kinds (at least in my fork, e.g. definitely for `Vec`) may not be the same as those returned by the existing `cast_*` free functions, which needs to be checked/fixed/have tests added for.

---

Minor naming nit: In the rest of `bytemuck`, there's usually a `cast_foo` (that succeeds or panics) and a `try_cast_foo` (that returns a `Result`), but `cast_in_place` returns a `Result`. I'd propose renaming to `try_cast_in_place` and having `cast_in_place` succeed or panic.

---

Since the `CastPtr` trait is where the alignment and type-safety checking happens, I believe there may need to be three different implementors, e.g. `SharedBorrowingPtr<T>`, `UniqueBorrowingPtr<T>`, `OwningPtr<T>`:

* `SharedBorrowingPtr`: (current `impl CastPtr for *const _`)
    * The container shared-borrows the value  (e.g. `&T`/`&[T]`).
    * Can relax or increase type alignment (as long as the actual pointer is validly aligned).
    * Source: `NoUninit`, Destination: `AnyBitPattern`, because no writes can occur as the destination type.

* `UniqueBorrowingPtr`: (current `impl CastPtr for *mut _`
    * The container uniquely borrows the value (e.g. `&mut T`/`&mut [T]`).
    * Can relax or increase type alignment (as long as the actual pointer is validly aligned).
    * Source and Destination: `NoUninit + AnyBitPattern`, because writes can occur as the destination type and be observed as the source type.

* `OwningPtr`: 
    * The container type owns the value.
    * Cannot relax or increase type alignment (allocations must be deallocated with the same layout (including alignment)). 
    * Source: `NoUninit`, Destination: `AnyBitPattern`, because either
        1. The container uniquely owns the value (e.g. `Box`), so writes can occur as the destionation type, but cannot be observed as the source type.
         2. The container shared-owns the value (e.g. `Rc`/`Arc`), so no writes can occur unless the container has guaranteed that it is uniquely owned, in which case the writes cannot be observed as the source type (Note: This would be relaxing bytemuck's current bounds. see also: https://github.com/Lokathor/bytemuck/pull/132#issuecomment-1234690814 . If this relaxation is not wanted, then another `CastPtr` type would be necessary).

---

I don't know that `Cow<T>` could work with this API. For `T: Sized`, it probably can't work because `Cow::Owned` owns a `T` directly, not behind a (castable) pointer. For `Cow<[T]>` it probably can't work (as-is, at least) because there's no place to put the `where T: Clone` bound needed to even make `Cow<[T]>` a well-formed type for `type Create<T: 'a>`.

src/cast.rs
```diff
@@ -0,0 +1,958 @@
...
+  unsafe fn from_raw(raw: Self::Raw) -> Self {
+    raw.map(|raw| C::from_raw(raw))
+  }
+}
+
+// SAFETY: `Pin` has no safety requirements for types which deref to an `Unpin` type.
+unsafe impl<'a, C> Container<'a> for Pin<C>
+where
+  C: Container<'a> + Deref<Target = <C as Container<'a>>::Item>,
+  C::Item: Unpin,
```
zachs18, line 328: Requiring `C::Target = C::Item` prevents casting things like `Pin<Box<[T]>>`, since `<Box<[T]> as Container>::Item = T`, but `<Box<[T]> as Deref>::Target = [T]`
```suggestion
  C: Container<'a> + Deref,
  C::Target: Unpin,
  C::Item: Unpin,
```

Jarcho:
Ended up rewriting this.

There's no longer a separation between casting the data and the metadata. It didn't really help with anything and added the complication of splitting and combining it in the first place. It's now been replaced with a single `CastRaw` trait.

The interface has been changed into a falliable/infalliable version. The infalliable version handles only the cases which can't fail and the falliable version handles all casts. This differs from how the current split works, but it's matches how `From`/`TryFrom` works and draws a more meaningful distinction between the two. A pair of traits for falliable/infalliable by value conversions was also added to match.

GAT's are no longer required, lowering the msrv to 1.57 (required for `panic!` in constants). We only ever need to refer to the two instances we already had as types so the lack of a type constructor can be worked around.

src/cast.rs
Jarcho: The change should go the other way with `Box<[T]>` setting the correct item type.

Jarcho:
>The container shared-owns the value (e.g. Rc/Arc), so no writes can occur unless the container has guaranteed that it is uniquely owned, in which case the writes cannot be observed as the source type (Note: This would be relaxing bytemuck's current bounds. see also:
https://github.com/Lokathor/bytemuck/pull/132#issuecomment-1234690814 . If this relaxation is not wanted, then another CastPtr type would be necessary).

By the sounds of the tracking issue it's likely to be the case that the bound can be relaxed. Still better to wait for that to be resolved.

src/cast.rs
```diff
@@ -0,0 +1,958 @@
...
+#[cfg(feature = "extern_crate_alloc")]
+impl<T, U> AssertClassContraints<T, U> for ArcWeakT {
+  const ASSERT: () = static_assert!(AssertSameAlign(T, U));
+}
+#[cfg(feature = "extern_crate_alloc")]
+impl<T, U> AssertClassContraints<T, U> for VecT {
+  const ASSERT: () = static_assert!(AssertSameAlign(T, U));
+}
+#[cfg(feature = "extern_crate_alloc")]
+impl<T: ToOwned, U: ToOwned> AssertClassContraints<T, U> for CowT {}
```
zachs18, line 906: Somwhere needs to assert/check that the source `T` and destination `U` of a casted `Cow<[T]>` -> `Cow<[U]>` have the same alignment, otherwise you can cast between `Cow::Owned`s and get undefined behavior:

```rs
// tests/unified_tests.rs
#[cfg(all(feature = "unified_cast", feature = "extern_crate_alloc"))]
#[test]
fn test_cow() {
  extern crate alloc;
  use alloc::{borrow::Cow, vec};
  let x: Cow<[u32]> = Cow::Borrowed(&[42]);
  // This correctly succeeds
  let y: Cow<[u8]> = x.reinterpret_inner();

  let x: Cow<[u32]> = vec![42].into();
  // This succeeds, but should fail since Vec<u32> cannot be cast to Vec<u8>.
  let y: Cow<[u8]> = x.reinterpret_inner(); 
}
```

```rs
$ cargo +nightly miri test
test test_cow ... error: Undefined Behavior: incorrect layout on deallocation: alloc71289 has size 4 and alignment 4, but gave size 4 and alignment 1
// ...
```

src/static_assert.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+  };
+}
+impl<T, U> AssertSizeMultipleOf<U> for T {}
+
+pub trait AssertSameAlign<T>: Sized {
+  const ASSERT: () = {
+    if align_of::<Self>() == align_of::<T>() {
+      panic!("Attempt to cast between two types with different alignments");
+    }
```
zachs18, line 65: This panics when the types have the same alignment, instead of when they have different alignments.
```suggestion
    if align_of::<Self>() != align_of::<T>() {
      panic!("Attempt to cast between two types with different alignments");
    }
```

src/static_assert.rs
```diff
@@ -0,0 +1,85 @@
...
+  };
+}
+impl<T> AssertNonZeroSize for T {}
+
+pub trait AssertSameSize<T>: Sized {
+  const ASSERT: () = {
+    if size_of::<Self>() == size_of::<T>() {
+      panic!("Attempt to cast between two types with different sizes");
+    }
```
zachs18, line 36: This panics when the types are the same size, instead of when they're different sizes.
```suggestion
    if size_of::<Self>() != size_of::<T>() {
      panic!("Attempt to cast between two types with different sizes");
    }
```

src/static_assert.rs
Jarcho: Looks like that's it for me tonight if I'm making those mistakes.

Jarcho:
Should be the last big change.  Added a `RawPointer` trait to normalize the differences between regular pointers and dst pointers.

This also adds `str` to the list of supported types. Though that is technically not defined, it must be layout compatible with `[u8]` for std's interface to work.

Lokathor:
I'm happy to see this making interesting progress, though I'll admit I haven't looked closely at the details yet.

Jarcho:
Fun thing about `panic` in associated constants, `cargo check` doesn't evaluate them. This means actually building the code is required for the errors to appear.

`trybuild` will need to be changed to use `cargo build` instead of `cargo check` to test for them.

Lokathor:
ah.

Probably this will get worse and worse with time, and rustc will have like..  `cargo check --consts` eventually or something. But until then a CI change is totally fine.

Jarcho:
Figured out how to make this work. An undocumented ~~feature~~ implementation detail of `trybuild` to get it to run `cargo build`, but it works.

The tests need to run as a separate crate in the same workspace. `trybuild` has a higher msrv than 1.36 and dev dependencies can't be optional.

Jarcho:
Should be in a working state now. A `uitest` workspace member was added and ci has a change to run it only on the stable channel. Hopefully the error messages don't change between versions too much, but diagnostic messages aren't stable from release to release.

Some more changes have been done to make sure the error message points to the call to `[try_]reinterpret[_inner]` and not somewhere in bytemuck's code.

Lokathor:
are we ready to transition into the review phase?

Jarcho:
I'll try to fix the error message for `ReinterpretInner` with incompatible types. It's currently a monstrosity that doesn't even point out the issue:

```
error[E0599]: the method `reinterpret_inner` exists for reference `&Padded`, but its trait bounds were not satisfied
  --> tests/ui/reinterpret_incompatible.rs:15:33
   |
7  | struct Padded(u16, u8);
   | -------------
   | |
   | doesn't satisfy `<_ as Container<'_>>::Class = _`
   | doesn't satisfy `Padded: ReinterpretInner<'_, _>`
   | doesn't satisfy `Padded: bytemuck::cast::Container<'_>`
   | doesn't satisfy `Padded: std::marker::Copy`
...
15 |   let _: &u32 = (&Padded(0, 0)).reinterpret_inner();
   |                                 ^^^^^^^^^^^^^^^^^ method cannot be called on `&Padded` due to unsatisfied trait bounds
   |
   = note: the following trait bounds were not satisfied:
           `<&Padded as bytemuck::cast::Container<'_>>::Class = _`
           which is required by `&Padded: ReinterpretInner<'_, _>`
           `&Padded: bytemuck::cast::Container<'_>`
           which is required by `&Padded: ReinterpretInner<'_, _>`
           `*const Padded: bytemuck::cast::RawPtr`
           which is required by `&Padded: ReinterpretInner<'_, _>`
           `Padded: std::marker::Copy`
           which is required by `&Padded: ReinterpretInner<'_, _>`
           `*const Padded: bytemuck::cast::CastRaw<_>`
           which is required by `&Padded: ReinterpretInner<'_, _>`
           `*const &Padded: bytemuck::cast::CastRaw<_>`
           which is required by `&&Padded: ReinterpretInner<'_, _>`
           `*mut &Padded: bytemuck::cast::CastRaw<_>`
           which is required by `&mut &Padded: ReinterpretInner<'_, _>`
           `<Padded as bytemuck::cast::Container<'_>>::Class = _`
           which is required by `Padded: ReinterpretInner<'_, _>`
           `Padded: bytemuck::cast::Container<'_>`
           which is required by `Padded: ReinterpretInner<'_, _>`
           `<&mut Padded as bytemuck::cast::Container<'_>>::Class = _`
           which is required by `&mut Padded: ReinterpretInner<'_, _>`
           `&mut Padded: bytemuck::cast::Container<'_>`
           which is required by `&mut Padded: ReinterpretInner<'_, _>`
           `*mut Padded: bytemuck::cast::RawPtr`
           which is required by `&mut Padded: ReinterpretInner<'_, _>`
           `Padded: std::marker::Copy`
           which is required by `&mut Padded: ReinterpretInner<'_, _>`
           `*mut Padded: bytemuck::cast::CastRaw<_>`
           which is required by `&mut Padded: ReinterpretInner<'_, _>`
note: the trait `bytemuck::cast::Container` must be implemented
  --> $WORKSPACE/src/cast.rs
   |
   | pub unsafe trait Container<'a>: Sized {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider annotating `Padded` with `#[derive(Clone, Copy)]`
    |
7   | #[derive(Clone, Copy)]
    |
```

Lokathor:
...once again the "note" on an error message is way too much until it's anti-helpful.

But sure, if you think you can somehow adjust it to make the error messages better please do.

Lokathor:
i'm closing this because it seems to have gone dead. if there's still interest from the author we could reopen

PR #193, opened 2023-06-05
Add functions for writing zeroed bytes to `&mut impl Zeroable` and `&mut [impl Zeroable]`

thomcc:
Two reasons this functionality is worth having:

1. Some cases in FFI want a struct required input to be cleared with `memset`, because they expect to add fields to the end of a struct and use `sizeof(TheStruct)` to indicate which fields exist on the type.

	That is: a new field could be added to the end of a struct without changing the size if the bytes of that field were previously tail padding -- in this case if all bytes were not zeroed in this manner, the uninit bytes in the field may be interpreted as having meaning.

2. `*blah = T::zeroed()` will end up having a few redundant copies of `T` on the stack in debug builds which could cause various problems.

I didn't add these as methods on `Zeroable` intentionally, but don't care much about it. If you'd prefer them there or something else I'm fine with it.

I'm pretty sure I didn't miss anything in terms of correctness here, although I had initially forgotten that `Zeroable` didn't imply `Copy`, so a bit of scrutiny wouldn't be out of place to ensure I'm not adding a problematic API.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -398,3 +398,33 @@ pub fn try_cast_slice_mut<
 ) -> Result<&mut [B], PodCastError> {
   unsafe { internal::try_cast_slice_mut(a) }
 }
+
+/// Fill all bytes of `target` with zeroes (see [`Zeroable`]).
+///
+/// This is similar to `*target = Zeroable::zeroed()`, but guarantees that any
+/// padding bytes in `target` are zeroed as well.
+///
+/// See also [`fill_zero`], if you have a slice rather than a single value.
+#[inline]
+pub fn write_zero<T: Zeroable + Copy>(target: &mut T) {
+  unsafe {
+    core::ptr::write_bytes(
+      target as *mut T as *mut u8,
+      0u8,
+      core::mem::size_of::<T>(),
+    )
+  }
+}
+
+/// Fill all bytes of `slice` with zeroes (see [`Zeroable`]).
+///
+/// This is similar to `slice.fill(Zeroable::zeroed())`, but guarantees that any
+/// padding bytes in `slice` are zeroed as well.
+///
+/// See also [`write_zero`], which zeroes all bytes of a single value rather
+/// than a slice.
+#[inline]
+pub fn fill_zero<T: Zeroable + Copy>(slice: &mut [T]) {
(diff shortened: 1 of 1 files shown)


thomcc:
I decided supporting non-copy types was actually useful. A lot of them won't need dropping and it isn't really that hard to handle the ones that do.

thomcc:
Also while this does clobber the provenance of any pointer it overwrites with zero bytes, that is true for writing `ptr::null()` too (as `null()` has no provenance).

(This is probably not news tho, I had just not realized it. Presumably this is part of why it's fine for raw pointers to have a `Zeroable` impl)

src/lib.rs
```diff
@@ -398,3 +398,41 @@ pub fn try_cast_slice_mut<
...
+#[inline]
+pub fn write_zero<T: Zeroable>(target: &mut T) {
+  unsafe {
+    core::ptr::drop_in_place(target);
+    core::ptr::write_bytes(
+      target as *mut T as *mut u8,
+      0u8,
+      core::mem::size_of::<T>(),
+    )
+  }
```
a1phyr, line 417: This will lead to a double-drop if `drop_in_place` panics. You should probably call `write_byte` in a drop guard instead.

Lokathor: This is true. Alternately, do we want to require Copy after all?

Lokathor: Alright, just checking i hadn't forgotten anything i guess

thomcc: hmmm, lets just make these require Copy

thomcc: Oh, sure, we can do it in the drop guard.

src/lib.rs
```diff
@@ -398,3 +398,51 @@ pub fn try_cast_slice_mut<
...
+  struct EnsureZeroWrite<T>(*mut T);
+  impl<T> Drop for EnsureZeroWrite<T> {
+    #[inline(always)]
+    fn drop(&mut self) {
+      unsafe {
+        core::ptr::write_bytes(
+          self.0.cast::<u8>(),
+          0u8,
+          core::mem::size_of::<T>(),
+        );
```
a1phyr, line 419:
```suggestion
        core::ptr::write_bytes(self.0, 0u8, 1);
```

PR #194, opened 2023-06-12
must_cast: fix test failures and fmt.

zachs18:
All doctests in `src/main.rs` were failing under miri due to https://github.com/rust-lang/miri/issues/2423. As a temporary workaround, instead mark those tests as `should_panic` under miri, and `compile_fail` only under not(miri) (b275f7dc76123854765e116e5085e65cdc43e69e).

Two doctests in `src/must.rs` were failing for unrelated reasons (4809de5b1294cb0106c6bbf7e09b5184803ea2de).

(Also `cargo fmt` on `src/must.rs`)

### DIFF
--- a/src/must.rs
+++ b/src/must.rs
@@ -81,8 +81,7 @@ pub fn must_cast_ref<A: NoUninit, B: AnyBitPattern>(a: &A) -> &B {
 /// let bytes: &mut [u8; 2] = bytemuck::must_cast_mut(&mut i);
 /// ```
 /// ```compile_fail,E0080
-/// # let mut bytes : [u8; 2] = [1, 2];
-/// # let bytes = &mut bytes[..];
+/// # let mut bytes: &mut [u8; 2] = &mut [1, 2];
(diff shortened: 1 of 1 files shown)


zachs18:
Shorter/clearer version of `cfg(not(miri))`-ing out the `compile_fail`s, adapted from @danielhenrymantilla via discord:
```rs
#[cfg(not(miri))]
macro_rules! compile_fail {() => (
    "``​`compile_fail,E0080"
)}

#[cfg(miri)]
macro_rules! compile_fail {() => (
    "``​`should_panic"
)}
```
at the beginning of `src/must.rs`, so as to:
```rs
#[doc = compile_fail!()]
```

(I'm not at a computer atm)

Lokathor:
Yeah i like the macro, please use that and also put in a comment to the miri issue as to why the macro is there at all.

zachs18:
I changed it to the macro version, and added a description of why the workaround is needed in a comment.

Issue #190, opened 2023-05-24
Allow deriving `Zeroable` with explicitly provided bounds


wdanilo:
Hi, I've got a code like this:

```
#[derive(Debug, Derivative, Zeroable)]
#[derivative(Default(bound = "T: Default"))]
pub struct LinkedCellArray<T, const N: usize = 32> {
    size:          Cell<usize>,
    first_segment: ZeroableOption<Box<Segment<T, N>>>,
}
```

And deriving `Zeroable` adds `T: Zeroable` bound. This is bad, as I've got `impl<T> Zeroable for ZeroableOption<T> {}` for any `T` implemented. Doing manual `impl<T, N> Zeroable for  LinkedCellArray<T, const N> {}` is also bad, as it does not check the soundness when the code is changed. Would it be possible to add an option to `Zeroable` to provide a custom bound, something like:

```
#[derive(Debug, Derivative, Zeroable)]
#[derivative(Default(bound = "T: Default"))]
#[zeroable(bound = "")]
pub struct LinkedCellArray<T, const N: usize = 32> {
    size:          Cell<usize>,
    first_segment: ZeroableOption<Box<Segment<T, N>>>,
}
```

Lokathor:
uh, possibly? I'm unclear how the derive code accepting a custom bound could be sure it's the correct bound.

Also, what's the ZeroableOption thing?

wdanilo:
The idea is that whatever is provided by the user will be used as the bound. If it is incorrect, the compiler will throw an error, so you, as an author of such macros, do not have to worry if the bound is correct. For example, consider the `Derivative` macro, which I can use as:

```rust
#[derive(Derivative)]
#[derivative(Clone(bound = ""))]
struct Foo<T> {
    t: T
}
```

The compiler will tell me that there is missing bound `T: Clone`. 

Regarding your second question, `ZeroableOption` is my custom option implementation that is zeroable:

```rust
#[repr(u8)]
pub enum ZeroableOption<T> {
    #[default]
    None = 0,
    Some(T),
}
```

Lokathor:
Well the compiler definitely will let you impl Zeroable when you have non-Zeroable fields. It's a unsafe trait with no methods, it *could* (in the "it will compile" sense) be implemented on anything. However, that's why it's unsafe, because the bounds of the impl must be correct for the type in question.

Given your specific ZeroableOption<T> type, i think perhaps you should just write down the unsafe impl directly.

[...]

PR #196, opened 2023-06-12
Zeroable derive custom bounds

zachs18:
Adds the ability to use custom bounds when using `derive(Zeroable)` *instead of* the normal "add a bound to each of the generics".

The normal soundness checks are still performed, so the custom bound must guarantee that the struct is actually zeroable, or compilation will fail.

Unresolved questions:

* Should bounds for each field implicitly be added (as described in https://github.com/Lokathor/bytemuck/issues/190#issuecomment-1563796078)? i.e. ["Perfect Derive"](https://smallcultfollowing.com/babysteps//blog/2022/04/12/implied-bounds-and-perfect-derive/#what-is-perfect-derive). This wouldn't be hard, and might make the whole "give explicit bounds" part of this PR obsolete (since the "bound on the fields" can be done without additional information), except as to *restrict* the impl.

<details>
<summary>Emit bounds for each field</summary>
If

```rs
#[derive(Zeroable)]
struct Struct<T, U, V> {
  a: Cell<usize>,
  b: PhantomData<T>,
  c: Option<Box<U>>,
  d: [V; 3],
}
```

emitted

```rs
unsafe impl<T, U, V> Zeroable for Struct<T, U, V>
  where
    Cell<usize>: Zeroable,
    PhantomData<T>: Zeroable,
    Option<Box<U>>: Zeroable,
    [V; 3]: Zeroable,
 {}
 ```
 
 that would work and not require the user to give any bounds.

</details>

Possible future work:
* Something similar for other traits (e.g. `Pod` for #191. I was looking at it, and it's harder than just re-using this PR's code because of the additional requirements of `Pod`.)

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -366,69 +366,61 @@ fn derive_marker_trait<Trait: Derivable>(input: DeriveInput) -> TokenStream {
     .unwrap_or_else(|err| err.into_compile_error())
 }
 
-/// Find a `#[name(key = "value")]` attribute on the struct, and parse "value"
-/// with `parser` and return it.
+/// Find `#[name(key = "value")]` helper attributes on the struct, and return
+/// their `"value"`s parsed with `parser`.
 ///
-/// Returns an error if multiple attributes with `name` are found, or if the one
-/// found does not match the expected format. Returns `Ok(None)` if no attribute
-/// with `name` is found.
-fn find_helper_attribute<P: syn::parse::Parser + Copy>(
+/// Returns an error any matching attributes do not match the expected format.
+/// Returns `Ok([])` if no attributes with `name` are found.
+fn find_helper_attributes<P: syn::parse::Parser + Copy>(
   attributes: &[syn::Attribute], name: &str, key: &str, parser: P,
-  example_value: &str,
-) -> Result<Vec<(syn::LitStr, P::Output)>> {
+  example_value: &str, invalid_value_msg: &str,
+) -> Result<Vec<P::Output>> {
+  let invalid_format_msg =
+    format!("{name} attribute must be `{name}({key} = \"{example_value}\")`",);
   let values_to_check = attributes.iter().filter_map(|attr| match &attr.meta {
-    syn::Meta::Path(path) => path.is_ident(name).then(|| {
-      Err(syn::Error::new_spanned(
-        &path,
-        format!(
-          "{name} attribute must be `{name}({key} = \"{example_value}\")`",
-        ),
-      ))
-    }),
+    // If a `Path` matches our `name`, return an error, else ignore it.
+    // e.g. `#[zeroable]`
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by wdanilo** [comment]
As I replied in the second thread, I believe the bounds generated for each field are a must-have, otherwise this is unsafe with wrongly provided bounds. If the user wants such an unsafe impl, they can always write `unsafe impl<...> Zeroable for ...`, but deriving should be always safe.

**Comment by zachs18** [comment]
As stated, the normal soundness checks are still performed, so the derive macro cannot produce unsound code that compiles, as it also emits checks that guarantee that each field is `Zeroable`. There is a doctest for this in the PR (lines 122-133 of derive/src/lib.rs).

<details>

<summary>For a concrete example</summary>

For a concrete example, this code (under this PR) expands to the following code.

```rs
#[derive(bytemuck::Zeroable)]
#[zeroable(bound = "u32: Copy, U: bytemuck::Zeroable")]
struct Asdf<T, U> {
    x: Box<T>,
    y: U,
}

// expands to
struct Asdf<T, U> {
    x: Box<T>,
    y: U,
}
const _: fn() = || {
    #[allow(clippy::missing_const_for_fn)]
    #[doc(hidden)]
    fn check<T, U>()
    where
        u32: Copy,
        U: bytemuck::Zeroable,
    {
        fn assert_impl<T: ::bytemuck::Zeroable>() {}
        assert_impl::<Box<T>>();
    }
};
const _: fn() = || {
    #[allow(clippy::missing_const_for_fn)]
    #[doc(hidden)]
    fn check<T, U>()
    where
        u32: Copy,
        U: bytemuck::Zeroable,
    {
        fn assert_impl<T: ::bytemuck::Zeroable>() {}
        assert_impl::<U>();
    }
};
unsafe impl<T, U> ::bytemuck::Zeroable for Asdf<T, U>
where
    u32: Copy,
    U: bytemuck::Zeroable,
{
}

```

The `fn check`s are where the soundness checks are performed, and they will not compile if the fields are not all guaranteed to be `Zeroable`.

(Also, note to self: the multiple separate `const _`s could probably be rolled into one.)

</details>

wdanilo:
@zachs18 what is the reason the expanded code generates the `const _: fn() = || { ... }` things? I don't think they are needed, unless I'm missing some obvious thing here.

zachs18:
(The failing CI on `mips`/`mips64` is due to https://github.com/rust-lang/rust/pull/113274 )
Okay, I added "perfect derive" semantics to `#[zeroable(bound = "...")]`, i.e. when explicit bounds are given, the given bounds are emitted in addition to a bound for each field's type. (If no explicit bounds are given, the current "bound each type generic" behavior is unchanged).

Marking this PR as ready for review @Lokathor @wdanilo .

Lokathor (approved):
Looks good to me, I'll wait a while for `wdanilo` to have a chance to look at it.

PR #200, opened 2023-07-19
Improve assembly of Contiguous derive

### LINKED ISSUE #175: Suboptimal assembly for Contiguous derive
The derived implementation of `Contiguous::into_integer` prevents some optimizations from happening. See the following example in which I have commented the resulting assembly as retrieved through [cargo-show-asm](https://github.com/pacak/cargo-show-asm) on my stable-x86_64-unknown-linux-gnu machine.

```rust
use bytemuck::Contiguous;

#[derive(Copy, Clone, Contiguous)]
#[repr(u8)]
pub enum E {
    E0,
    E1,
    E2,
}

impl E {
    // movzx eax, byte ptr [rdi]
    pub fn ref_as(&self) -> Option<Self> {
        Self::from_integer(*self as u8)
    }

    // movzx eax, byte ptr [rdi]
    pub fn ref_into(&self) -> Option<Self> {
        Self::from_integer(self.into_integer())
    }

    // mov eax, edi
    pub fn val_as(self) -> Option<Self> {
        Self::from_integer(self as u8)
    }

    // cmp dil, 3
    // mov eax, 3
    // cmovb eax, edi
    pub fn val_into(self) -> Option<Self> {
        Self::from_integer(self.into_integer())
    }
}
```

The assembly for `E::val_into` should be the same as for `E::val_as`.

This could be an issue with the default forbidden-to-override implementation of `Contiguous::into_integer` or it could be an issue with rustc. As a workaround the derive macro could emit `as Int` since it knows what the repr type is. If you feel this is a rustc issue then I am happy to report it there.

e00E:
`from_integer` and `into_integer` are usually provided by the trait's default implementation. We override this implementation because it goes through `transmute_copy`, which can lead to inefficient assembly as seen in https://github.com/Lokathor/bytemuck/issues/175 .

Closes #175

Implemented my suggestion from when I reported the issue. The downside of this approach is that we're duplicating code between the derive and the default implementation. The upside is the better assembly from the derived implementation. It can do better than the default implementation because it has more information.

I changed the range check to `contains` because Clippy complained about about it and I wanted both version of the function to have the same check.
Edit: Reverted because MSRV doesn't support `contains`.

I verified that the linked issue is fixed by this PR.

### DIFF
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -371,12 +371,31 @@ impl Derivable for Contiguous {
     let min_lit = LitInt::new(&format!("{}", min), input.span());
     let max_lit = LitInt::new(&format!("{}", max), input.span());
 
+    // `from_integer` and `into_integer` are usually provided by the trait's default implementation.
+    // We override this implementation because it goes through `transmute_copy`, which can lead to
+    // inefficient assembly as seen in https://github.com/Lokathor/bytemuck/issues/175 .
+
     Ok((
       quote!(),
       quote! {
           type Int = #integer_ty;
           const MIN_VALUE: #integer_ty = #min_lit;
           const MAX_VALUE: #integer_ty = #max_lit;
+
+          #[inline]
+          fn from_integer(value: Self::Int) -> Option<Self> {
+            #[allow(clippy::manual_range_contains)]
+            if Self::MIN_VALUE <= value && value <= Self::MAX_VALUE {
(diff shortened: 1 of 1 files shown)


e00E:
I see CI fails because `contains` isn't stable on the MSRV. Will change it back to manual if.

Lokathor:
Sorry this got lost for a while, but it looks great!

PR #204, opened 2023-08-14
Fix derive non-run doctests and expand docs for Pod and TransparentWrapper derives.

zachs18:
Doctests only run in the crate proper; doctests inside of integration tests are not run. Move the doctests from the integration test `derive/tests/basic.rs` to be doctests on their respective derive macro defintions in `derive/lib.rs` so they are actually run, expanding them[^2][^3] to make them better for documentation as well.

Also updated the documentation for the `Pod` and `TransparentWrapper` derive macros to reflect their current state (`derive(Pod)` allows generics if and only if `#[repr(transparent)]`; `derive(TransparentWrapper)` requires any ZST fields be `Zeroable`).

---

[^2]: `NewtypeWrapperTest` was testing that `#[derive(Pod, ...)] #[repr(transparent)] struct Generic<A>(A);` is only `Pod` if `A: Pod`, which is now shown by lines 61-88 on the docs for `Pod`'s derive macro.
[^3]: `TransarentWrapperZstTest` was testing that `#[derive(TransparentWrapper)] struct Wrapper(T, NonZeroableZST);` fails, which is now shown by lines 305-320 on the docs for `TransparentWrapper`'s derive macro.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -23,19 +23,69 @@ use crate::traits::{
 /// - All fields in the struct must implement `Pod`
 /// - The struct must be `#[repr(C)]` or `#[repr(transparent)]`
 /// - The struct must not contain any padding bytes
-/// - The struct contains no generic parameters
+/// - The struct contains no generic parameters, if it is not
+///   `#[repr(transparent)]`
 ///
-/// ## Example
+/// ## Examples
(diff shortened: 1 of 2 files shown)


Lokathor:
I love when tests actually get run!

PR #205, opened 2023-08-29
Add Unaligned wrapper/utility type

HeroicKatora:
This utility wrapper allows casting into the _representation_ of a `T` without the need for the alignment to match. In some applications, such as zerocopy networking or some filesystem internals, it is an important property to check the proper size and encode the intentended type semantics of a raw bytestream. In particular, the relevant data may be payload within a larger bytestream and consequently not aligned. This intermediate representation helps delay the copy into an aligned value until proper checks have been performed (including performing copies of parts of the payload between those two states).

The type is not trivial as usually, access to an _unaligned_ field of a `packed` struct is `unsafe`. The type offers only a safe API since `bytes_of` can be used for other needs as easily. If future usecases arise it can be extended.

The type should live here or in `core` but this is easier to motivate :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,6 +130,9 @@ pub use offset_of::*;
 mod transparent;
 pub use transparent::*;
 
+mod unaligned;
+pub use unaligned::*;
+
 #[cfg(feature = "derive")]
 pub use bytemuck_derive::{
   AnyBitPattern, ByteEq, ByteHash, CheckedBitPattern, Contiguous, NoUninit,
--- a/src/unaligned.rs
+++ b/src/unaligned.rs
@@ -1,0 +1,182 @@
+use crate::{Pod, Zeroable};
+
+/// A wrapper around an unaligned value with the bit-pattern of a `T`.
+///
+/// The type is modeled along the lines of its sibling, the [`Cell`](#core::cell::Cell). These are
+/// similar in that the types disallow reference-access to their contents while preserving the
+/// types' bit-representation. However, this types constraints come from a different reason. Its
+/// trade-off is that the field containing the bit representation of the value is not aligned
+/// properly. It is therefore not possible to create references to the inner value, either. Yet
+/// some operations on the value are still sound and this type encapsulates those unsafe primitives
+/// into a safe API.
+///
+/// The `Unaligned` container does not interact with `Send` and `Sync` and does not permit updates
+/// through shared references. 
+///
+/// # Example
+///
+/// One primary use for this type is to use for attribute in larger POD structs, such as network
+/// packets.
+///
+/// ```
+/// use bytemuck::{Pod, Zeroable, Unaligned, from_bytes};
+///
+/// # unsafe impl Zeroable for Icmp {}
+/// # unsafe impl Pod for Icmp {}
+///
+/// /// A structure describing the layout of an ICMPv4 header.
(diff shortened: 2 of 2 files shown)


Lokathor:
I said most or all of this on the discord, but I want to capture it here as well:

I'm not convinced that this wrapper type is the correct way to model multi-byte data that might be unaligned. I think instead of having `Icmp` have a field that's a `u16` (and thus align2) you should use a type like what [pack1](https://docs.rs/pack1/latest/pack1/) offers, where the type is internally an array of however many `u8` (and thus align1), and then methods to turn data into and out of that type. Such a type can also force an endianness in a much clearer way, which is obviously critical for data formats between systems.

In general, I think that `repr(packed)` was probably a mistake in the language design that we shouldn't continue.

If there really is an absolute need for this wrapper type then it could have a place in `bytemuck`, but I'd like to hear concretely why something like `pack1` isn't appropriate to the use cases this is trying to cover.

HeroicKatora:
Another crate that came up with the type itself independently (it leans primarily into the `Cell` parallels). Links:

* https://github.com/kvverti/unaligned
* https://crates.io/crates/unaligned

----

The reason for making it a wrapper instead of duplicating between primitive types is that it actually is a very generic nod to Rust's memory model, and not something types need to consider when they are defined. Values are really just their bit-representation and the primitive just affirms that. I don't consider `repr(packed)` a bigger mistake than the initial version of `uninit`, the value it adds to the language in completing the toolbox available within the typesystem is very valuable; however it is barebones and does not effectively interact at the current state.

The example is a little short for motivating the generic wrapper, certainly. The advantage comes from the fact that one does not _need_ to write separate structs for the on-wire and the by value in-memory representation. In particular the use-case also include structs with highly aligned SIMD attributes, or zerocopy serialization of types defined in other crates. It's not necessary to split these into separate types keeping the type usage precise and discoverable. The advantage of using a separate byte array type for integers is of course the reminder of platform-dependence however I think those two effects _should_ be treated separately especially since it is possible.

The main problem in the opposite, using the unaligned type throughout, is a vague concern of performance / codegen. I have no evidence of the effect but it certainly removes some alignment information from the optimizer inputs.

However, in consideration of the separate crate, and that a different way to defined the type for _unsized_ `T` exists at the cost of not dropping the value¹, it seems a bit rash anyways to immediately decide on the PR. I did not compare against alternatives comprehensively.

----

<details>
<summary>Initial thoughts on alternative formulation I've not yet explored in detail</summary>

Rust will always generate `Drop`-glue, and `Drop::drop(&mut self)` makes it necessary to move the inner field for the dropping of the inner value. However, unsized fields _can not_ be moved in that fashion and hence they are not allowed in packed structs. Except, of course they are if we opt into _not dropping the field_ by marking it with `ManuallyDrop`.

```rust

#[repr(packed)]
struct Unaligned<T: ?Sized> {
    inner: ManuallyDrop<T>,
}
```

The cost here is that the field is _never_ dropped (writing a manual `Drop` impl runs into the same problems as the auto-generated one and we can't / don't want to specialize on dropping sized types by `ManuallyDrop::take`). That might be okay, the primary purpose is to have the type behind a reference as a result of `from_bytes` etc. But it warrants caution.

</details>

Lokathor:
I'm gonna close this for inactivity.

If there's a stronger motivation that comes up later I'm willing to revisit the topic.

PR #206, opened 2023-08-30
Use safe, public APIs where possible to remove some `unsafe` blocks.

anforowicz:
`trait CheckedBitPattern` constraints the associated `type Bits` as follows: `Bits: AnyBitPattern`.  This means that `checked::try_from_bytes` doesn't need to call `unsafe` `internal::try_from_bytes` but can instead call the (safe) public API `crate::try_from_bytes` (thanks to the `AnyBitPattern` constraint).  Similar reasoning can be applied to the other snippets modified by this PR.  Being able to successfully compile the crate after this PR seems to confirm this reasoning.

### DIFF
--- a/src/checked.rs
+++ b/src/checked.rs
@@ -243,7 +243,7 @@ impl From<crate::PodCastError> for CheckedCastError {
 pub fn try_from_bytes<T: CheckedBitPattern>(
   s: &[u8],
 ) -> Result<&T, CheckedCastError> {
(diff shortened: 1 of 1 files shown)


Lokathor:
This all looks good. If someone can report a specific performance problem we could go back to directly calling the unsafe internal versions, but otherwise it's probably good policy to go through the safe interface when possible.

PR #203, opened 2023-07-30
improve documentation for optional features

joseluis:
I realized the documentation in docs.rs could benefit by explicitly indicating the items and implementations that depends on which specific features. Please let me know if you want anything done differently. This is the summary of the changes so far:

- new feature `nightly_docs` enables `doc_cfg` unstable feature that allows showing explicit dependencies on features.
- new cargo alias `nightly_docs` for a more convenient local documentation build. E.g.: `cargo +nightly nightly_docs --open`.
- annotate exported items and implementations depending on some feature with the corresponding `doc(cfg())` attribute.
- add the `nightly_docs` feature to the `[package.metadata.docs.rs]` section.

PR #207, opened 2023-09-05
Have `cfg`s as part of the input to `impl_unsafe_marker_for_simd`

zachs18:
... so that the macro can correctly expand with `doc(cfg(...))` attributes under `feature = "nightly_docs"`.

(see https://github.com/Lokathor/bytemuck/pull/203#issuecomment-1707208974 )


zachs18:
Ah, `cargo fmt`s snuck in. I can remove those if wanted.

Lokathor:
probably better to have the formatting applied eventually, might as well be now.

PR #209, opened 2023-09-06
Added custom crate path for bytemuck with tests

AndriBaal:
Adds the option to specify a custom path for the bytemuck crate, instead of `::bytemuck`. This allows the bytemuck macros to be reexported.

### DIFF
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -21,17 +21,17 @@ macro_rules! bail {
 }
 
 pub trait Derivable {
-  fn ident(input: &DeriveInput) -> Result<syn::Path>;
-  fn implies_trait() -> Option<TokenStream> {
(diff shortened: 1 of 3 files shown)


derive/src/traits.rs
```diff
@@ -1219,3 +1219,47 @@ mod tests {
...
+
+  let mut crate_name = quote!(::bytemuck);
+  for attr in &input.attrs {
+    if !attr.path().is_ident("bytemuck") {
+      continue;
+    }
+
+    attr.parse_nested_meta(|meta| {
+      if meta.path.is_ident(ATTR_NAME) {
+        let expr: syn::Expr = meta.value()?.parse()?;
```
fu5ha, line 1234: You should be able to just parse directly to a `syn::LitStr` here, no?

fu5ha (commented):
Thanks! Overall lgtm, I think the attribute parsing logic can be simplified a bit though.

PR #211, opened 2023-10-12
Add from_box_bytes and box_bytes_of with BoxBytes type

### LINKED ISSUE #167: Getting allocated bytes (Box/Rc/Arc) out of Box<T>

ia0:


src/allocation.rs
```diff
@@ -686,4 +687,105 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>:
...
+}
+
+impl Drop for BoxBytes {
+  fn drop(&mut self) {
+    // SAFETY: See type invariant.
+    unsafe { alloc::alloc::dealloc(self.ptr.as_ptr(), self.layout) };
+  }
+}
+
+impl<T> From<Box<T>> for BoxBytes {
```
zachs18, line 729:
```suggestion
impl<T: NoUninit> From<Box<T>> for BoxBytes {
```

This impl will be publicly accessible since `BoxBytes` is a public type, so it must have the required bounds to be sound in user-accessible code.

src/allocation.rs
```diff
@@ -686,4 +687,105 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>:
...
+    let ptr = Box::into_raw(value) as *mut u8;
+    // SAFETY: Box::into_raw() returns a non-null pointer.
+    let ptr = unsafe { NonNull::new_unchecked(ptr) };
+    BoxBytes { ptr, layout }
+  }
+}
+
+/// Re-interprets `Box<T>` as `BoxBytes`.
+#[inline]
+pub fn box_bytes_of<T: NoUninit + AnyBitPattern>(input: Box<T>) -> BoxBytes {
```
zachs18, line 741:
```suggestion
pub fn box_bytes_of<T: NoUninit>(input: Box<T>) -> BoxBytes {
```

Converting from `T` to `[u8]` only requires `T: NoUninit`

src/allocation.rs
```diff
@@ -686,4 +687,105 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>:
...
+) -> Result<Box<T>, (PodCastError, BoxBytes)> {
+  let layout = Layout::new::<T>();
+  if input.layout.align() != layout.align() {
+    return Err((PodCastError::AlignmentMismatch, input));
+  } else if input.layout.size() != layout.size() {
+    return Err((PodCastError::SizeMismatch, input));
+  } else {
+    // SAFETY: See type invariant.
+    Ok(unsafe { Box::from_raw(input.ptr.as_ptr() as *mut T) })
+  }
```
zachs18, line 774: `input` must be wrapped in `ManuallyDrop`, otherwise there is a use-after-free when `input` is dropped at the end of the function.
```suggestion
  } else {
    let input = ManuallyDrop::new(input);
    // SAFETY: See type invariant.
    Ok(unsafe { Box::from_raw(input.ptr.as_ptr() as *mut T) })
  }
```

src/allocation.rs
```diff
@@ -686,4 +687,105 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>:
...
 
-impl<I: ?Sized, T: ?Sized + TransparentWrapper<I>> TransparentWrapperAlloc<I> for T {}
+impl<I: ?Sized, T: ?Sized + TransparentWrapper<I>> TransparentWrapperAlloc<I>
+  for T
+{
+}
+
+/// As `Box<[u8]>`, but remembers the original alignment.
+pub struct BoxBytes {
+  // SAFETY: `ptr` is owned and was allocated with `layout`.
```
zachs18, line 697:
```suggestion
  // SAFETY: `ptr` is owned, was allocated with `layout`, and points to `layout.size()` initialized bytes.
```

src/allocation.rs
```diff
@@ -686,4 +687,105 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>:
...
+    Ok(unsafe { Box::from_raw(input.ptr.as_ptr() as *mut T) })
+  }
+}
+
+impl BoxBytes {
+  /// Constructs a `BoxBytes` from its raw parts.
+  ///
+  /// # Safety
+  ///
+  /// The pointer is owned and has been allocated with the provided layout.
```
zachs18, line 782:
```suggestion
  /// The pointer is owned, has been allocated with the provided layout, and points to `layout.size()` initialized bytes.
```

src/allocation.rs
```diff
@@ -686,4 +687,105 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>:
...
+}
+
+/// Re-interprets `BoxBytes` as `Box<T>`.
+///
+/// ## Panics
+///
+/// * If the input isn't aligned for the new type
+/// * If the input's length isn’t exactly the size of the new type
+#[inline]
+pub fn try_from_box_bytes<T: NoUninit + AnyBitPattern>(
```
zachs18, line 763:
```suggestion
pub fn try_from_box_bytes<T: AnyBitPattern>(
```

zachs18 (changes requested):
The concept looks fine to me. The invariant should also mention that the bytes may not be uninitialized. The bound on the `From` impl must be changed to be sound, and the bounds on `box_bytes_of`/`(try_)from_box_bytes` bounds should be relaxed.

src/allocation.rs
zachs18: Alternately, add a `BoxBytes::into_raw_parts` method, and use that here.

Lokathor:
I'm gonna rub the CI and let zachs take the lead on the reviewing for now. The basic idea seems reasonable

src/allocation.rs
ia0: Good point, I've introduced `into_raw_parts`.

ia0 (commented):
Thanks a lot for the quick review! I guess the next step is to add tests if the change looks acceptable.

ia0:
I've removed the fact that it fixes #167 from the description because it doesn't fix it in its generality. Only the `T: Sized` case.

zachs18 (changes requested):
Looks good to me! (with `from_raw_parts`)

I think these functions can be extended to allow `[T]` later in a semver-compatible way (using a [sealed trait](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) bound), so I think it'd be fine to merge this as-is and add that functionality later.

<details>
 <summary>Making these functions accept `[T]`</summary>

```rs
// Note: pub items in non-pub module cannot be named
// outside the crate unless re-exported.
// These traits are intentionally not re-exported publicly,
// as user implementations would be incorrect.
// See https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed
mod sealed {
  use crate::{BoxBytes, PodCastError};
  use alloc::boxed::Box;

  pub trait BoxBytesOf {
    fn box_bytes_of(self: Box<Self>) -> BoxBytes;
  }
  pub trait FromBoxBytes {
    fn try_from_box_bytes(
      bytes: BoxBytes,
    ) -> Result<Box<Self>, (PodCastError, BoxBytes)>;
  }
}
impl<T: NoUninit> sealed::BoxBytesOf for T { ... }
impl<T: NoUninit> sealed::BoxBytesOf for [T] { ... }

impl<T: AnyBitPattern> sealed::FromBoxBytes for T { ... }
impl<T: AnyBitPattern> sealed::FromBoxBytes for [T] { ... }

pub fn box_bytes_of<T: sealed::BoxBytesOf>(input: Box<T>) -> BoxBytes {
  input.box_bytes_of()
}
// etc
```

This would be semver-compatible (as a [minor](https://doc.rust-lang.org/cargo/reference/semver.html#fn-generalize-compatible) change) with the current defintions (it relaxes a trait bound), but would need some mention in the docs of what types are allowed, since the sealed-trait trick makes the generated docs slightly unclear.

</details>

src/allocation.rs
```diff
@@ -686,4 +687,117 @@ pub trait TransparentWrapperAlloc<Inner: ?Sized>:
...
+}
+
+impl BoxBytes {
+  /// Constructs a `BoxBytes` from its raw parts.
+  ///
+  /// # Safety
+  ///
+  /// The pointer is owned, has been allocated with the provided layout, and
+  /// points to `layout.size()` initialized bytes.
+  pub unsafe fn from_parts(ptr: NonNull<u8>, layout: Layout) -> Self {
```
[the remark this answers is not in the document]
ia0: Oops, that was my intention. Thanks for catching that!

zachs18 (approved):

Lokathor (approved):

Lokathor:
I'm going to approve and merge this right now, but I'll hold off on publishing for a little bit. I'm not 100% convinced that the function names are all the best, given that this type is a little unusual, so I'll just think about it over the weekend.

PR #220, opened 2024-01-24
Explicitly document that pod_read_unaligned and try_pod_read_unaligned don't panic on unallowed reads

hniksic:
Their names make that reasonably clear already, but it doesn't hurt to be explicit. Also, when choosing between `*from_bytes(x)` and `pod_read_unaligned(x)`, it's good to have a clearly documented criterion.

The PR also adds a couple other cosmetic doc fixes, as separate commits.


Lokathor (approved):

Lokathor:
released in `bytemuck-1.14.1`

PR #228, opened 2024-03-02
Convert `BoxBytes` to/from boxed slices

zachs18:
As I mentioned [in a comment on the BoxBytes PR](https://github.com/Lokathor/bytemuck/pull/211#pullrequestreview-1675256320), `BoxBytes` logically can be converted to/from boxed slices as well as boxed `Sized` types. This PR extends the existing `box_bytes_of`/`(try_)from_box_bytes` using sealed traits to allow them to take/return boxed slices as well as their current boxed Sized types (and `Box<str>` for `box_bytes_of`).

<details> <summary>semver</summary>

The relaxation of the function bounds is not a breaking change.

The change from `impl<T: NoUninit> From<Box<T>> for BoxBytes` to `impl<T: ?Sized + BoxBytesOf> From<Box<T>> for BoxBytes` I believe is not a breaking change:

* Any type for which a downstream crate was previously allowed to implement `From<Box<U>> for BoxBytes` must have `U` as a local, non-`NoUninit` type, which as far as I can tell is still allowed under this new impl.
</details>

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -9,6 +9,8 @@
 //!   `bytemuck = { version = "VERSION_YOU_ARE_USING", features =
 //! ["extern_crate_alloc"]}`
 
+use self::sealed::BoxBytesOf;
+
 use super::*;
 #[cfg(target_has_atomic = "ptr")]
 use alloc::sync::Arc;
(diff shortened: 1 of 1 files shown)


Lokathor (approved):

Lokathor:
Released `bytemuck-1.15.0`

Please report any compatibility problems that somehow appear.

PR #243, opened 2024-05-24
Address `unexpected_cfgs` warnings instead of allowing the lint

Urgau:
r? @Lokathor

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,9 @@ nightly_docs = []
 [dependencies]
 bytemuck_derive = { version = "1.4", path = "derive", optional = true }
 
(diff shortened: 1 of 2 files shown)


Urgau:
If your interested in knowing more, [here](https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html) is the Cargo specific doc and [here](https://doc.rust-lang.org/nightly/rustc/check-cfg.html) the `rustc --check-cfg` doc.

The output of the lint links directly to the Cargo specific doc.

> To the uninitiated it doesn't read well.

If you have ideas on how to improve the situation I'm happy to here them.

PR #249, opened 2024-06-06
Don't require T to be AnyBitPattern for MaybeUninit<T> to be AnyBitPattern

tyilo:
`MaybeUninit<T>` is always valid for any bit pattern even if `T` is not.

### DIFF
--- a/src/anybitpattern.rs
+++ b/src/anybitpattern.rs
@@ -60,5 +60,6 @@ unsafe impl<T: Pod> AnyBitPattern for T {}
   feature = "nightly_docs",
   doc(cfg(feature = "zeroable_maybe_uninit"))
 )]
-unsafe impl<T> AnyBitPattern for core::mem::MaybeUninit<T> where T: AnyBitPattern
+unsafe impl<T> AnyBitPattern for core::mem::MaybeUninit<T> where
+  T: Copy + 'static
 {}


zachs18:
This was going to be implemented in #152, but was discovered to possibly be unsound in general (https://github.com/Lokathor/bytemuck/pull/152#issuecomment-1336252308), so a never-unsound-but-more-conservative implementation was implemented instead (#160, require T to be `AnyBitPattern` so we know it has no interior mutability).If [`Freeze`](https://github.com/rust-lang/rust/issues/121675) was stable, then this would be possible (either `impl<T: Freeze> AnyBitPattern for MaybeUninit<T>`, or under #153, `impl<T> AnyBitPattern for MaybeUninit<T>`).

tyilo:
Ah, so this is only sound if implementing `Copy` means a type also implements `Freeze`.

As I understand it, there are currently no `!Freeze` type that implements `Copy` so this change would be sound for the current version of Rust.

However this could change in a future version of Rust, where this change would then be unsound.

Lokathor:
I've heard occasional rumblings of wanting Freeze to be a a stable trait some day. It wouldn't be any time soon, but more people are coming around to it being quite useful. So, I think that we should wait until that happens.

Lokathor:
I'm going to close this until Rust as a language develops further, and then maybe someday later we can revisit the concept.

Issue #253, opened 2024-07-20
Zero-sized type (ZST) handling inconsistency between borrowed and owned slice casts.


zachs18:
(For the purposes of this issue, assume `Src`/`Dst`/`Zst`/`NonZst` have the same alignment. Borrowed and owned casts intentionally have different alignment requirements, which are not the issue here.)

`bytemuck::try_cast_slice<Src, Dst>`[^2] [is documented as](https://docs.rs/bytemuck/1.16.1/bytemuck/fn.try_cast_slice.html) returning an error (specifically `Err(SizeMismatch)`) when `Src` and `Dst` are a ZST and a non-ZST (in either order).

However, `bytemuck::allocation::try_cast_slice_box<Src, Dst>`[^1]  says nothing in [its documentation](https://docs.rs/bytemuck/1.16.1/bytemuck/allocation/fn.try_cast_slice_box.html) regarding ZSTs, currently always returns `Ok(empty)` when `Src` is a ZST and `Dst` is not, and currently *panics* with a modulo-by-zero error when `Src` is a non-ZST and `Dst` is a ZST. ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0cda1505fa555be47eb31149dcaeacb0))

I see three possible overall resolutions to the inconsistency:

1. Keep (and document?) the inconsistency (but fix the panic).
    * `try_cast_slice_box` allows `[Zst] -> [NonZst]`, returning an empty slice.
    * `try_cast_slice` disallows `[Zst] -> [NonZst]` and `[NonZst] -> [Zst]`, returning `Err(SizeMismatch)`
    * Status quo
2. Make `cast_slice` behave like `cast_slice_box` (but fix the panic).
    * `cast_slice` and `cast_slice_box` allow `[Zst] -> [NonZst]` casts, returning an empty slice.
    * May break code relying on `cast_slice::<Zst, NonZst>` not succeeding.
3. Make `cast_slice_box` behave like `cast_slice`.
    * Disallow any casting between non-ZST <-> ZST slices, returning `Err(SizeMismatch)`.
    * May break code relying on `cast_vec::<Zst, NonZst>` succeeding.
    * (Implies option 2 below)
    
and two possible resolutions for the `[NonZst] -> [Zst]` panic:

1. Allow empty source slices for `[NonZst] -> [Zst]`, return an error[^3] on non-empty source slices.
    * This matches the existing documentation that the start and end content size must be the same.
[...]

PR #254, opened 2024-07-21
Fix panics in `try_cast_slice_box` etc

zachs18:
Addresses the panics described in #253, but does not change the other inconsistencies (i.e. option 1 overall "status quo" + option 1 for the panic "Allow empty source slices for `[NonZst] -> [Zst]`, return an error on non-empty source slices").


Lokathor (approved):
Regardless of what we decide in 253, we definitely want to remove these panic possibilities.

PR #255, opened 2024-07-21
Disallow casting ZST <-> non-ZST in cast_slice_box.

zachs18:
Alternative to #254. Fixes the panics in `try_cast_slice_box` by disallowing casting between boxed ZST slices and boxed non-ZST slices, regardless of length (i.e. option 3 from #253 "Make cast_slice_box behave like cast_slice").

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -21,7 +21,7 @@ use alloc::{
   vec::Vec,
 };
 use core::{
-  mem::ManuallyDrop,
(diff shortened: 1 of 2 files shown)


Lokathor:
I merged 254 instead of this because i think that we should always be trying to move towards more casts allowed than less. That said, I won't actually publish the patch right away, I'd like to think on 253 a bit more and then resolve that fully and then do a release with a "complete" solution.

PR #257, opened 2024-07-28
derive(Zeroable) on fieldful enums and repr(C) enums

### LINKED ISSUE #230: Allow deriving `Zeroable` for enums which are `#[repr(u32, i8, etc..)]`
If an enum is explicitly defined in a manner where "0" is a valid variant, and all the payloads of the 0 variant are zeroable, then it concludes that the enum is safely zeroable.

zachs18:
Closes #230.

Allows deriving `Zeroable` on `enum`s where:

1. The enum is `#[repr(Int)]`, `#[repr(C)]`, or `#[repr(C, Int)]`,
2. There is a variant with discriminant 0,
3. The variant with discriminant 0 is either fieldless, or all of its fields are `Zeroable`.

Also allows using the "perfect derive with additional bounds" from #196 for Zeroable on enums, where the fields for the "perfect derive" are from the variant with discriminant 0 (see the `MyOption` example on line 189 of derive/src/lib.rs).

Internal changes:

* `VariantDiscriminantIter` now also gives the `&'a Variant` associated with the discriminant.
* the comment on line 338 of derive/src/traits.rs kept getting mangled by `cargo fmt`, so I put it on its own line.
* In `derive_marker_trait_inner`: `let predicates = ...` moved to after `let fields = ...`, since computing `fields` now requires accessing `input`, but `predicates` borrows it mutably. Does not change semantics.
* `generate_fields_are_trait` and some other helper functions now take an `Option<&Variant>` to operate on if they otherwise don't support enums.
* some `cargo fmt`
* drive-by clarification/typo fixes (e.g. renaming `discriminantor` -> `discriminant`)

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -116,7 +116,7 @@ pub fn derive_anybitpattern(
 
 /// Derive the `Zeroable` trait for a type.
 ///
-/// The macro ensures that the struct follows all the the safety requirements
+/// The macro ensures that the type follows all the the safety requirements
 /// for the `Zeroable` trait.
 ///
 /// The following constraints need to be satisfied for the macro to succeed on a
@@ -133,7 +133,7 @@ pub fn derive_anybitpattern(
 /// - All fields in the variant with discriminant 0 (if any) must implement
 ///   `Zeroable`
 ///
-/// The macro always succeeds on `union`s.
+/// The macro always succeeds on unions.
 ///
 /// ## Example
 ///
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -6,7 +6,7 @@ mod traits;
 
 use proc_macro2::TokenStream;
 use quote::quote;
-use syn::{parse_macro_input, DeriveInput, Result};
+use syn::{parse_macro_input, DeriveInput, Result, Variant};
 
 use crate::traits::{
   bytemuck_crate_name, AnyBitPattern, CheckedBitPattern, Contiguous, Derivable,
@@ -114,14 +114,26 @@ pub fn derive_anybitpattern(
   proc_macro::TokenStream::from(expanded)
 }
 
-/// Derive the `Zeroable` trait for a struct
+/// Derive the `Zeroable` trait for a type.
 ///
 /// The macro ensures that the struct follows all the the safety requirements
 /// for the `Zeroable` trait.
 ///
-/// The following constraints need to be satisfied for the macro to succeed
+/// The following constraints need to be satisfied for the macro to succeed on a
+/// struct:
 ///
-/// - All fields in the struct must to implement `Zeroable`
+/// - All fields in the struct must implement `Zeroable`
+///
+/// The following constraints need to be satisfied for the macro to succeed on
+/// an enum:
+///
+/// - The enum has an explicit `#[repr(Int)]`, `#[repr(C)]`, or `#[repr(C,
+///   Int)]`.
+/// - The enum has a variant with discriminant 0 (explicitly or implicitly).
+/// - All fields in the variant with discriminant 0 (if any) must implement
+///   `Zeroable`
+///
+/// The macro always succeeds on `union`s.
 ///
 /// ## Example
 ///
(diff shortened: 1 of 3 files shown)


derive/src/traits.rs
```diff
@@ -906,7 +939,7 @@ fn generate_checked_bit_pattern_enum_with_fields(
 fn generate_assert_no_padding(input: &DeriveInput) -> Result<TokenStream> {
   let struct_type = &input.ident;
   let span = input.ident.span();
-  let fields = get_fields(input)?;
+  let fields = get_fields(input, None)?;
```
danielhenrymantilla, line 968:
```suggestion
  let enum_variant = None; // `no padding` check is not supported for `enum`s yet.
  let fields = get_fields(input, enum_variant)?;
```

derive/src/traits.rs
```diff
@@ -1215,7 +1248,7 @@ impl<'a, I: Iterator<Item = &'a Variant> + 'a> Iterator
       self.last_value += 1;
```
danielhenrymantilla, line 1274: ⚠️ Unrelated to this PR, but it seems rather important (somebody compiling proc-macros with a different choice of the `overflow-checks` knob may run into a panic with this logic):

```suggestion
      self.last_value = self.last_value.wrapping_add(1);
```

derive/src/traits.rs
```diff
@@ -202,6 +205,24 @@ impl Derivable for Zeroable {
...
+      Data::Enum(DataEnum { variants, .. }) => {
+        let iter = VariantDiscriminantIterator::new(variants.iter());
+        for res in iter {
+          match res {
+            Ok((0, variant)) => return Some(variant.fields.clone()),
+            Ok(_) => (),
+            Err(_) => return None,
+          }
+        }
+        None
```
danielhenrymantilla, line 226: Third occurrence of this pattern; it may warrant a helper function 🤔

derive/src/lib.rs
```diff
@@ -133,7 +133,7 @@ pub fn derive_anybitpattern(
 /// - All fields in the variant with discriminant 0 (if any) must implement
 ///   `Zeroable`
 ///
-/// The macro always succeeds on `union`s.
+/// The macro always succeeds on unions.
```
danielhenrymantilla: Debatable change; I personally prefer to use Rust keywords whenever possible since those necessarily have a clear and unambiguous meaning.

derive/src/traits.rs
```diff
@@ -44,6 +44,11 @@ pub trait Derivable {
   fn explicit_bounds_attribute_name() -> Option<&'static str> {
     None
   }
+
+  /// If this trait has a custom meaning for "perfect derive", this function
+  /// should be overridden to return `Some`.
+  ///
+  /// The default is "the fields of a struct; unions and enums not supported".
```
danielhenrymantilla, line 51: Good and important doc 👌 

  - 🤔 I wonder if we shouldn't be moving said default logic over here (by changing the output to `::syn::Result<Fields>`) to ensure actual code logic and this doc comment remain in sync

danielhenrymantilla (approved):
LGTM, nice! ✅

derive/src/traits.rs
zachs18: Hmm, in the case of the (unstable) `repr128` feature, this could cause unsoundness with `derive(CheckedBitPattern)` on an `enum X { A = i64::MAX as i128, B }`, so to be safe, I'll make `last_value` an `i128` and then do the `wrapping_add` (since the compiler will catch any overflows at the actual discriminant type).

derive/src/traits.rs
zachs18: Latest push adds a helper function `get_zero_variant(enum: &DataEnum) -> Option<&Variant>`

fu5ha (approved):
Nice, LGTM. I'm deferring the in depth soundness checking of the macro to @danielhenrymantilla 's review whom I trust.

Lokathor:
Sorry about the delay, `bytemuck_derive-1.8.0` should be out now

PR #258, opened 2024-07-29
Fix UB dropping BoxBytes with a zero-sized layout

zachs18:
`Box` does not actually perform an allocation for zero-sized values, so `BoxBytes::drop` should not call `dealloc` if `self.layout.size() == 0`.

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -742,8 +742,9 @@ impl<I: ?Sized, T: ?Sized + TransparentWrapper<I>> TransparentWrapperAlloc<I>
 
 /// As `Box<[u8]>`, but remembers the original alignment.
 pub struct BoxBytes {
-  // SAFETY: `ptr` is owned, was allocated with `layout`, and points to
-  // `layout.size()` initialized bytes.
+  // SAFETY: `ptr` is owned, points to `layout.size()` initialized bytes, and
+  // was allocated with `layout` (unless `layout.size() == 0` in which case it
+  // is dangling).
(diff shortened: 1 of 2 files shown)


src/allocation.rs
```diff
@@ -742,8 +742,9 @@ impl<I: ?Sized, T: ?Sized + TransparentWrapper<I>> TransparentWrapperAlloc<I>
 
 /// As `Box<[u8]>`, but remembers the original alignment.
 pub struct BoxBytes {
-  // SAFETY: `ptr` is owned, was allocated with `layout`, and points to
-  // `layout.size()` initialized bytes.
+  // SAFETY: `ptr` is owned, points to `layout.size()` initialized bytes, and
+  // was allocated with `layout` (unless `layout.size() == 0` in which case it
+  // is dangling).
```
kupiakos, line 747: Note: this terminology is incorrect [according to Rust's definition](https://github.com/rust-lang/reference/blob/master/src/behavior-considered-undefined.md#dangling-pointers). A pointer to zero-sized data is _never_ dangling, since [all zero-sized reads and writes are now considered valid](https://github.com/rust-lang/unsafe-code-guidelines/issues/472), even with a null pointer.

The primary issue is that `dealloc` requires that the layout not be zero-sized; whether the pointer dangles is irrelevant.

Lokathor: That is definitely true in a strict sense, and a follow up PR for fixing that note would be appropriate if you wanted to do that, because safety notes should not be loose with wording.

PR #259, opened 2024-07-29
Add `bytemuck` attribute to `NoUninit` derive

richardpringle:
Not sure if you didn't add this attribute here for a reason, but it works perfectly for re-exporting the crate when you do

(I can also add the attribute to the other `proc_macro_derive`s if you'd like).


Lokathor:
Normally I'd let fu5ha make all final approvals on proc-macro changes, but since this is already used in other places within the derives crate I'll just merge this small change now.

richardpringle:
Thanks @Lokathor (for both the merge and the crate)!

PR #261, opened 2024-08-08
Make must_cast by-value and by-shared-ref functions const

zachs18:
(but not by-mut-ref, since that's [not stable](https://github.com/rust-lang/rust/issues/57349)).

This would raise the MSRV of the `must_cast` feature from 1.57.0 to 1.64.0 (when `slice_from_raw_parts` was made const-stable). Alternately, this could be added as a new feature (`const_must_cast`?) with its own MSRV[^1]. 

The new arm of the `transmute!` macro is to work around borrowing a generic type in const-eval not being allowed on stable, since the borrowed type could have interior mutabilty. Transmuting via `union` does not have this issue, since no borrowing occurs.

[^1]: `must_cast` and `must_cast_ref`'s MSRV is at least 1.61.0, when [`const_fn_trait_bound`](https://github.com/rust-lang/rust/pull/93827/) was stabilized. `must_cast_slice`'s MSRV is at least 1.64.0, as mentioned.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ min_const_generics = [] # MSRV 1.51: support arrays via min_const_generics
 wasm_simd = []    # MSRV 1.54.0: support wasm simd types
 aarch64_simd = [] # MSRV 1.59.0: support aarch64 simd types
 
-must_cast = [] # MSRV 1.57.0: support the `must` module.
+must_cast = [] # MSRV 1.64.0: support the `must` module.
 
 const_zeroed = [] # MSRV 1.75.0: support const `zeroed()`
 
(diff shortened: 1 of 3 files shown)


zachs18:
(I believe the CI failures are due to https://github.com/rust-lang/rust/issues/128808)

Lokathor:
Because it's an optional feature and because 1.64 is quite old at this point i think we can ignore the msrv.

I'd like to probably wait on CI if the nightly rust bug is going to only be a day or two

zachs18:
(Current nightly has the fix; force-pushed for CI)

Lokathor:
Released `bytemuck-v1.17.0`

PR #268, opened 2024-08-27
Force `#[repr(C)]` layout to guarantee same offset of `union` fields.

anforowicz:
https://rust-lang.github.io/unsafe-code-guidelines/layout/unions.html points out that

    > [...] the default layout of Rust unions is, in general,
    > unspecified.
    >
    > That is, there are no general guarantees about the offset of the
    > fields, whether all fields have the same offset, what the call ABI
    > of the union is, etc.

This commit explicitly asks for `#[repr(C)]` layout to guarantee that both fields have the same offeset.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,6 +127,7 @@ macro_rules! transmute {
   // since the compiler hedges that the type being borrowed could have interior mutability.
   ($srcty:ty; $dstty:ty; $val:expr) => {
     {
+      #[repr(C)]
       union Transmute<A, B> {
         src: ::core::mem::ManuallyDrop<A>,
         dst: ::core::mem::ManuallyDrop<B>,


anforowicz:
This commit is based on the `cargo vet` audit being discussed at https://chromium-review.googlesource.com/c/chromium/src/+/5813463/6/third_party/rust/chromium_crates_io/vendor/bytemuck-1.17.0/src/lib.rs

Lokathor:
The offset of the field doesn't matter *in this case* because we're not accessing the field via any sort of pointer offset math, but instead by using the standard field access syntax.

That said, this doesn't hurt, so whatever.

Lokathor:
released in `bytemuck-1.17.1`

anforowicz:
> released in `bytemuck-1.17.1`

Thanks!  Much appreciated.

> The offset of the field doesn't matter _in this case_ because we're not accessing the field via any sort of pointer offset math, but instead by using the standard field access syntax.

If the offsets were different, then I don't think it would matter that no pointer offset math was used.  I think that this would be equivalent to expecting that transmuting can work through something like:

```
// `A` and `B` are `AnyBitPattern` and `NoUninit`
type A = u32;  
type B = i32;

// The compiler can _theorerically_ insert some `_padding` bytes before `A` or `B`.
// (Since there are no guarantees for the default non-`#[repr(C)]` layout.)
#[repr(C)]
struct AWithPadding {
    _padding: [u8; 4],  // 4 bytes of padding
    a: A,
}
#[repr(C)]
struct BWithPadding {
    _padding: [u8; 8],  // Different padding - 8 bytes
    b: B,
}

// Now `u.a_with_padding.a` and `u.b_with_padding.b` don't use pointer offset math,
// but `A` and `B` don't "overlap" in `ABUnion` and therefore the `union` can't be used for transmuting.
#[repr(C)]
union ABUnion {
    a_with_padding: AWithPadding,
    b_with_padding: BWithPadding,
}
```

> That said, this doesn't hurt, so whatever.

Ack.  I think that in practice there will be no padding (since A and B have the same size), but since there are no guarantees, it's probably best to explicitly request offset 0 by using `#[repr(C)]`.

PR #279, opened 2024-10-19
Remove incorrect spanned quote usage

GnomedDev:
`quote_spanned` should be used for associating `Box::new(#init)` with the `init` span, for example. This crate is associating a lot of the returned tokens with the input tokens, which breaks clippy's "are these tokens from a proc macro" checks, which leads to issues like https://github.com/rust-lang/rust-clippy/issues/13349 which are unsolvable from our end (to my knowledge).

This PR swaps with normal quote, and adds a test to make sure that none of the `allow` attributes get associated with input tokens to prevent the specific clippy issue from reappearing.

### DIFF
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -2,7 +2,7 @@
 use std::{cmp, convert::TryFrom};
 
 use proc_macro2::{Ident, Span, TokenStream, TokenTree};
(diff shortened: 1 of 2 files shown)


danielhenrymantilla (approved):

Lokathor:
may take me a bit to publish this but thanks for the fix

PR #283, opened 2024-11-05
Add functions to allocate zeroed `Arc` and `Rc`

a1phyr:
These require rustc 1.82

### DIFF
--- a/src/allocation.rs
+++ b/src/allocation.rs
@@ -149,6 +149,38 @@ pub fn zeroed_slice_box<T: Zeroable>(length: usize) -> Box<[T]> {
   try_zeroed_slice_box(length).unwrap()
 }
 
+/// Allocates a `Arc<T>` with all contents being zeroed out.
+#[cfg(all(feature = "alloc_uninit", target_has_atomic = "ptr"))]
+pub fn zeroed_arc<T: Zeroable>() -> Arc<T> {
+  let mut arc = Arc::new_uninit();
+  crate::write_zeroes(Arc::get_mut(&mut arc).unwrap()); // unwrap never fails for a newly allocated Arc
+  unsafe { arc.assume_init() }
+}
+
+/// Allocates a `Arc<[T]>` with all contents being zeroed out.
+#[cfg(all(feature = "alloc_uninit", target_has_atomic = "ptr"))]
+pub fn zeroed_arc_slice<T: Zeroable>(length: usize) -> Arc<[T]> {
+  let mut arc = Arc::new_uninit_slice(length);
+  crate::fill_zeroes(Arc::get_mut(&mut arc).unwrap()); // unwrap never fails for a newly allocated Arc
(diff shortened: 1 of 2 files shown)


src/allocation.rs
```diff
@@ -149,6 +149,42 @@ pub fn zeroed_slice_box<T: Zeroable>(length: usize) -> Box<[T]> {
   try_zeroed_slice_box(length).unwrap()
 }
 
+/// Allocates a `Arc<T>` with all contents being zeroed out.
+pub fn zeroed_arc<T: Zeroable>() -> Arc<T> {
```
zachs18, line 153:
```suggestion
/// Allocates a `Arc<T>` with all contents being zeroed out.
#[cfg(target_has_atomic = "ptr")]
pub fn zeroed_arc<T: Zeroable>() -> Arc<T> {
```

src/allocation.rs
```diff
@@ -149,6 +149,42 @@ pub fn zeroed_slice_box<T: Zeroable>(length: usize) -> Box<[T]> {
...
+  unsafe {
+    let arc = Arc::new_uninit();
+    core::ptr::write_bytes(Arc::as_ptr(&arc) as *mut T, 0, 1);
+    arc.assume_init()
+  }
+}
+
+/// Allocates a `Arc<[T]>` with all contents being zeroed out.
+pub fn zeroed_arc_slice<T: Zeroable>(length: usize) -> Arc<[T]> {
```
zachs18, line 162:
```suggestion
/// Allocates a `Arc<[T]>` with all contents being zeroed out.
#[cfg(target_has_atomic = "ptr")]
pub fn zeroed_arc_slice<T: Zeroable>(length: usize) -> Arc<[T]> {
```

zachs18 (changes requested):
so that this will still compile on targets without `Arc`

zachs18:
Also: It is currently up in the air if it is valid to write to the pointer returned from `{Arc,Rc}::as_ptr`: https://github.com/rust-lang/rust/issues/87862 https://github.com/rust-lang/rust/pull/104337 , so ideally these would use `{Arc,Rc}::get_mut(&mut arc).unwrap()` at least until that is clarified[^1].

[^1]: The compiler can optimize out the `unwrap` panic branch for `Rc` but not `Arc`, but it should still never be possible to panic at runtime since the `Arc` is newly allocated

src/allocation.rs
a1phyr: I also forgot `cfg(feature = "alloc_uninit")`, surprised that the CI didn't catch it

a1phyr:
As you wish. I updated the code to use `write_zeroes` and `fill_zeroes`, it is cleaner like that if we use `get_mut`

Lokathor (approved):

Lokathor:
released in bytemuck-1.20

PR #284, opened 2024-11-06
Improve `fill_zeroes`' use of `ptr::write_bytes`

a1phyr:


Lokathor:
Looks like these are deliberately two lines to prevent the length check (using `&self`) from invalidating the mut pointer.

Lokathor (approved):

Lokathor:
released in bytemuck-1.20

PR #285, opened 2024-11-27
Implements `Pod` and `Zeroable` for AVX512 types that don't require nightly. Additionally allows for the utilization of `must_cast_mut` and `must_cast_slice_mut` in const-contexts on Rust 1.83+

herabit:
Title is long, but, you get the picture.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
-* Implement `Pod` and `Zeroable` for `__m512`, `__m512d` and `__m512i` without nightly.
-  Requires Rust 1.70, and is gated through the `avx512_simd` cargo feature.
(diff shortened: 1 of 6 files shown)


changelog.md
```diff
@@ -1,5 +1,16 @@
 # `bytemuck` changelog
 
+## Unreleased
+
+* Implement `Pod` and `Zeroable` for `__m512`, `__m512d` and `__m512i` without nightly.
+  Requires Rust 1.70, and is gated through the `avx512_simd` cargo feature.
```
zachs18, line 6:
```suggestion
* Implement `Pod` and `Zeroable` for `core::arch::{x86, x86_64}::__m512`, `__m512d` and `__m512i` without nightly.
  Requires Rust 1.72, and is gated through the `avx512_simd` cargo feature.
```

Lokathor: I don't know that we need the prefix part, but the correctly listed msrv is essential

Lokathor:
released in `bytemuck-1.21.0`

PR #291, opened 2024-12-28
fix deriving CheckedBitpattern on packed struct

Freax13:
This PR fixes the code emitted by the `CheckedBitPattern` derive macro to make it compatible with packed structs. It also adds a regression test for this.


Freax13:
AFAICT the failing CI check is unrelated to this PR (it also fails on main).

stegaBOB (approved):
This looks good to me. Its essentially the same thing I did, but copies the values out for all structs including non packed ones. I don't think it really matters much for a debug implementation.

stegaBOB:
Hey @Lokathor would it be possible to publish a patch bump for this for the derive crate?

Lokathor:
yes, I'll see if i can do it via phone, otherwise it'll have to wait a few hours

Lokathor:
1.9.1 should be out now

also, the previously published 1.9.0, which had been yanked at the request of some folks using an old rust toolchain, is now unyanked

PR #307, opened 2025-03-19
Remove rust-version and resolver from derive/Cargo.toml

### LINKED ISSUE #306: edition2024 required for bytemuck_derive 1.9.1
Hey there!

I am using anchor, which uses Cargo 1.79.0, and I kept running into an issue with `bytemuck_derive`.

It defaults to 1.9.1, but when I build, it fails with 

```
 [...]
  failed to parse manifest at `/usr/src/app/vendor/bytemuck_derive/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized
 in this version of Cargo (1.79.0 (8d3d6ef09 2024-08-15)).
```

If I patch it to use `bytemuck_derive-v1.8.1`, this does not occur.

If you're here looking for a workaround, add 

```
[patch.crates-io]
bytemuck_derive = { git = "https://github.com/Lokathor/bytemuck", tag = "bytemuck_derive-v1.8.1" }
```

to your anchor/solana Cargo.toml,

or, if you can change the dependencies of the programs themselves, you can add
```
[dependencies]
bytemuck_derive = ">=1.8.1, <1.9.0"
```
to the crate's Cargo.toml

Tritlo:
This fixes #306, and bumps the version.

The root cause is a bit of a SNAFU: `bytemuck` specifies a version dependency of `1.4.1` on `bytemuck_derive`, which should be checked and fetched. Since this expands to `>=1.4.1, <2.0.0`, it happily accepts `1.9.1` even if it can't be compiled with a less than 6 month old compiler version.

```
Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized
 in this version of Cargo (1.81.0 (2dbb1af80 2024-08-20)).
```

By removing these fields which are non-essential (as of yet), we retain backwards compatibility. Once we *require* a later version of Rust, we can add these fields accordingly, and bump the major version so that an older, backwards compatible package is chosen.

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "bytemuck_derive"
 description = "derive proc-macros for `bytemuck`"
-version = "1.9.1"
+version = "1.9.2"
 authors = ["Lokathor <zefria@gmail.com>"]
 repository = "https://github.com/Lokathor/bytemuck"
 readme = "README.md"
@@ -9,8 +9,6 @@ keywords = ["transmute", "bytes", "casting"]
 categories = ["encoding", "no-std"]
 edition = "2018"
 license = "Zlib OR Apache-2.0 OR MIT"
-rust-version = "1.84"
-resolver = "3"
 
 [lib]
 name = "bytemuck_derive"


Lokathor:
there's a few points I want clear up for the record.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

The `path` is only used for local development. When building from the registry, only versions in the registry are considered. So the `path` entry won't ever affect anyone who isn't directly working on stuff in this repository.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements

The version "1.4.1" actually expands to ">=1.4.1, <2.0.0", so even if the `path` were removed, cargo would naturally attempt to select the latest published version less than "2.0.0".

---

Because there's a work around, as discussed in the linked issue, and because the `rust-version` key has to be specified in as many published versions as possible for cargo to be able to select a correct version for use during version selection, I'm afraid I can't accept this PR. The `rust-version` key will never actually be strictly necessary for the crate to operate, but we're using it now that it's part of stable rust.

It's unfortunate to have to use a work-around, but this is just how it's gonna be. Hopefully you can get your toolchains updated to 1.84 or later some time soon.

Tritlo:
seems pretty weird to have your package broken for a huge ecosystem for no gain at all. Irresponsible even.

Tritlo:
1.84 got released barely 70 days ago, so any downstream dependency that hasn't been updated in the last two months won't compile anymore.

You should seriously consider giving downstream organizations and ecosystem a longer grace period.

Lokathor:
which-version-of-rust-do-you-use 1

---

That said, this is fixed in `bytemuck_derive-1.9.2` by removing `resolver` without needing to remove `rust-version`.

Tritlo:
This poll doesn't cover downstream users of tools that bundle a rust compiler, and shows that 42% were using 1.83, which does not support edition 2024. 

But thank you for the fix, much appreciated 🙏

Lokathor:
Well 1.84 wasn't even released as a Stable version during the survey period (last December), at the time it would have been the Beta. When it says 90% use "latest Stable", it's the case that most people update their rust toolchain on the day of release.

I try to have one of the most permissive MSRV policies I know of, but I'm still shocked every time someone *actually* has an old Rust.

PR #309, opened 2025-03-22
Extend #308 with `-unwind` support

chorman0773:
Adding to #308, this adds `C-unwind` and `system-unwind` fn-ptr `ZeroableInPod` impls behind the `zeroable_unwind_fn` feature gate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,9 @@ track_caller = []
 # MSRV 1.74.0 Pod/Zeroable implementations for `core::num::Saturating`
 pod_saturating = []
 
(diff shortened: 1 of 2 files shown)


Lokathor:
the 1.34 check didn't pass, is the cfg somehow not working?

chorman0773:
> the 1.34 check didn't pass, is the cfg somehow not working?

Yep. Apparently `extern` tags get validated when parsing before Rust 1.41, instead of after macro/cfg expansion, so you can't even gate an extern tag on `cfg` in those versions (or use an attribute macro either, I'd assume).

PR #310, opened 2025-03-26
Set rust version for `bytemuck_derive` to 1.61

nicoburns:
~~https://github.com/Lokathor/bytemuck/pull/293/commits/7ae54e291b8b3d4a89ee4a199ae693412cb6b50f sets the edition to `2018` so that bytemuck_derive can build with Rust `1.34`.~~ `bytemuck_derive` will build with Rust `1.61`,  however, the `rust-version` field in `Cargo.toml` is set to `1.84` not `1.61`. We should set the version to `1.61` so that users with older rust versions can compile the crate.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,6 +19,7 @@ jobs:
         include:
         # versions (all on linux-x86_64)
         - { rust: 1.34.0, os: ubuntu-latest }
+        - { rust: 1.61.0, os: ubuntu-latest }
         - { rust: stable, os: ubuntu-latest }
         - { rust: beta, os: ubuntu-latest }
         - { rust: nightly, os: ubuntu-latest }
@@ -36,6 +37,8 @@ jobs:
     - run: cargo test --verbose
     - run: cargo test --verbose --no-default-features
     - run: cargo test --verbose
+    - run: cargo test --verbose --features derive
+      if: matrix.rust == '1.61.0'
     - run: cargo test --verbose --all-features
       if: matrix.rust == 'nightly'
     - run: cargo test --verbose --manifest-path=derive/Cargo.toml --all-features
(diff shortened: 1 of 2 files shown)


Lokathor:
bytemuck_derive definitely can't build on 1.34

Lokathor:
https://github.com/Lokathor/bytemuck/blob/main/.github/workflows/rust.yml#L39
yeah here we go, the all-features check, including the derives, are only checked on nightly.

nicoburns:
It actually builds on Rust 1.61 in my testing (the reason it won't build on 1.60 being that the latest version of `syn` requires 1.61). I'll update the PR to that.

Lokathor:
I think it doesn't matter before 1.84, because resolvers before 1.84 don't consider rust-version when selecting package versions.

separately, we actually should probably add a CI run that checks the derive crate on the version we say it builds on. not that we'd hold back the minimum version of the derives, but that we'd update the cargo.toml and CI combo whenever that run fails.

nicoburns:
> I think it doesn't matter before 1.84, because resolvers before 1.84 don't consider rust-version when selecting package versions.

I believe it's the other way around. It matters more for versions before 1.84. Versions from 1.84 onwards have the ability to automatically select an older version of `bytemuck_derive` if the latest version has too new of an MSRV. Older versions of rustc will just fail to compile if the `rust-version` is too new:

Output from `cargo +1.61 build` on `main` below:
```bash
❯ cargo +1.61 build
warning: unused manifest key: lints
error: package `bytemuck_derive v1.9.2 (/Users/nico/code/oss/bytemuck/derive)` cannot be built because it requires rustc 1.84 or newer, while the currently active rustc version is 1.61.0
```

(with this PR the same code compiles just fine with `1.61`)

Lokathor:
I published this, though i accidentally used `publish` and not `release` and it went through anyway since the PR updated the version, which means there's no autogenerated git tag and stuff, but that's probably fine

nicoburns:
Thanks! You should be able to manually create a git tag if you want to:

```bash
git tag -a bytemuck_derive-v1.9.3 -m "chore: Release bytemuck_derive version 1.9.3"
git push --tags
```

PR #311, opened 2025-04-04
Impl core's Error trait on Rust 1.81+

joshuamegnauth54:
Hello awesome bytemuck devs!

Rust's `Error` trait is available in `core` as of 1.81. This small patch adds a crate feature to derive [core::error::Error](https://doc.rust-lang.org/nightly/core/error/trait.Error.html) instead of bringing in `std`.

I added an extra feature instead of changing the original derive so that it doesn't affect users on older Rust versions. I also accounted for the case when both features are enabled for some reason.


Lokathor:
I don't really use the Error trait myself, but I am wondering if we should override the information methods to give better information, but that's a concern for another day.

PR #317, opened 2025-07-16
impl `AnyBitPattern` for `[MaybeUninit<T: AnyBitPattern>; N]`

zdivelbiss:
Provides the following implementation:

```rust
#[cfg(all(feature = "zeroable_maybe_uninit", feature = "min_const_generics"))]
#[cfg_attr(
  feature = "nightly_docs",
  doc(cfg(all(
    feature = "zeroable_maybe_uninit",
    feature = "min_const_generics"
  )))
)]
unsafe impl<T, const N: usize> AnyBitPattern for [core::mem::MaybeUninit<T>; N] where
  T: AnyBitPattern
{
}
```

Which allows arrays of `MaybeUninit<T: AnyBitPattern>` to be treated as `AnyBitPattern`. For instance, the following struct is not currently possible via the safer derive macro for `AnyBitPattern`:

```rust
#[repr(align(0x10))]
#[derive(AnyBitPattern)]
struct Stack<const N: usize>([MaybeUninit<u8>; N]);
```

Which results in the following error on the struct's tuple field:
```
the trait bound `core::mem::MaybeUninit<u8>: bytemuck::Pod` is not satisfied
```

... which complicates my current use case, of trying to safely allocate a `Stack` of uninitialized bytes (skipping the cost of first zeroing the stack, which isn't required for a process stack).

### DIFF
--- a/src/anybitpattern.rs
+++ b/src/anybitpattern.rs
@@ -62,3 +62,16 @@ unsafe impl<T: Pod> AnyBitPattern for T {}
 )]
 unsafe impl<T> AnyBitPattern for core::mem::MaybeUninit<T> where T: AnyBitPattern
 {}
+
+#[cfg(all(feature = "zeroable_maybe_uninit", feature = "min_const_generics"))]
+#[cfg_attr(
+  feature = "nightly_docs",
+  doc(cfg(all(
+    feature = "zeroable_maybe_uninit",
+    feature = "min_const_generics"
+  )))
+)]
+unsafe impl<T, const N: usize> AnyBitPattern for [core::mem::MaybeUninit<T>; N] where
+  T: AnyBitPattern
+{
+}


fu5ha (commented):
This makes sense to me, but I think we could actually broaden this to instead be

```rust
#[cfg(feature = "min_const_generics")]
#[cfg_attr(
  feature = "nightly_docs",
  doc(cfg(
    feature = "min_const_generics"
  ))
)]
unsafe impl<T, const N: usize> AnyBitPattern for [T; N] where
  T: AnyBitPattern
{
}
```

Which would then get `[MaybeUninit<T>; N]` for free.

Also this is making me think that we could broaden the existing `impl<T> AnyBitPattern for MaybeUninit<T> where T: AnyBitPattern` to drop the `T: AnyBitPattern` bound. `MaybeUninit<T>` is in general `AnyBitPattern` no matter what is the inner `T`, unless I'm missing something 🤔

zachs18:
> This makes sense to me, but I think we could actually broaden this to instead be
> 
> ```rust
> #[cfg(feature = "min_const_generics")]
> #[cfg_attr(
>   feature = "nightly_docs",
>   doc(cfg(
>     feature = "min_const_generics"
>   ))
> )]
> unsafe impl<T, const N: usize> AnyBitPattern for [T; N] where
>   T: AnyBitPattern
> {
> }
> ```

This would conflict with the existing `impl<T: Pod> Pod for [T; N]` (due to the blanket `impl<T: Pod> AnyBitPattern for T`).

In a future (major) version of `bytemuck`, we could "flip" the blanket impl, and have `impl<T: AnyBitPattern + NoUninit> Pod for T`, and have users only implement `AnyBitPattern` and `NoUninit`, and then this would work (since we would have `impl<T: AnyBitPattern> AnyBitPattern for [T; _]` and `impl<T: NoUninit> NoUninit for [T; _]` which would together mean `T: Pod` implies `[T; _]: Pod`)


> Also this is making me think that we could broaden the existing `impl<T> AnyBitPattern for MaybeUninit<T> where T: AnyBitPattern` to drop the `T: AnyBitPattern` bound. `MaybeUninit<T>` is in general `AnyBitPattern` no matter what is the inner `T`, unless I'm missing something 🤔

This has possible issues with interior mutability. In particular, if `T` has interior mutability, then so does `MaybeUninit<T>`, and `AnyBitPattern` currently requires that the type not have interior mutability, so `impl<T> AnyBitPattern for MaybeUninit<T>` would be unsound. We could restrict it to `impl<T: Copy> AnyBitPattern for MaybeUninit<T>` since there are *currently* no interior-mutable `Copy` types, but it is not clear if `T: Copy` is _guaranteed_ to never have interior mutability. More info here: https://github.com/Lokathor/bytemuck/pull/249#issuecomment-2151995040

fu5ha (approved):

PR #322, opened 2025-08-19
chore: use new stable avx512 types from rust 1.89

nsarlin-zama:
With the release of rust 1.89, `stdarch_x86_avx512` has been stabilized and do not need the use of a nightly toolchain.
This PR adds support for these types in the `avx512_simd` feature. 

The `nightly_stdsimd` feature has been kept for backward compatibility but I can remove it if you prefer.

### DIFF
--- a/src/pod.rs
+++ b/src/pod.rs
@@ -170,14 +170,14 @@ impl_unsafe_marker_for_simd!(
 );
 
 impl_unsafe_marker_for_simd!(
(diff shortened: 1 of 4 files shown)


src/pod.rs
zachs18, line 1: These should also use `all(.., any(..))`, like in zeroable.rs, not just `all(..)`

PR #333, opened 2025-11-07
Remove `extern "stdcall"` fn ptr impls on non-x86-32 windows.

### LINKED ISSUE #332: `extern "stdcall" fn` implements `ZeroableInOption` on all Windows targets, but `extern "stdcall"` is only valid on x86-32 Windows.
https://github.com/rust-lang/rust/issues/137018 tracks a future-compatibility warning added in Rust 1.89 for disallowing `extern "stdcall"` on non-x86-32 Windows targets.

`bytemuck` currently implements `ZeroableInOption` for `extern "stdcall"` fn pointers on all Windows targets, not just x86-32 Windows. See https://github.com/Lokathor/bytemuck/actions/runs/19155827739/job/54756069053?pr=331#step:6:8 for an example of the FCW being hit in bytemuck's x86_64-pc-windows-msvc CI job.

Some possible approaches:

1. Only have these impls on x86-32 Windows. (i.e. remove the impls on non-x86-32 Windows) https://github.com/zachs18/bytemuck/commit/77e9fb907541ba04564e0e080ba804170fee4902
   ```diff
   -#[cfg(target_os = "windows")]
   +#[cfg(all(target_os = "windows", target_arch = "x86"))]
   ```
   1. I think any *direct* dependant that would be affected by this removal would already be hitting the FCW in their own crate, but since the FCW is IIUC not yet warn-in-deps, *indirect* dependants that could be affected by this might not know yet. I don't know enough about the Windows Rust ecosystem to judge how likely this is.
2. Move these impls behind a temporary feature on non-x86-32 Windows, so users who still need them can enable the feature until they update their code. https://github.com/zachs18/bytemuck/commit/9f32c80746d7ef08684bf8edd6eca95140a8a502
   ```diff
   -#[cfg(target_os = "windows")]
   +#[cfg(all(target_os="windows", any(target_arch = "x86", feature = "fn_ptr_unsupported_abi")))]
   ```
    1. this feature could be made enabled-by-default to avoid breaking dependants, at least until the FCW progresses to warn-in-deps.
3. Wait for the FCW to progress to warn-in-deps.

@Lokathor Which of these approaches seems better to you, or something else?

zachs18:
Closes #332.

### DIFF
--- a/src/zeroable_in_option.rs
+++ b/src/zeroable_in_option.rs
@@ -52,9 +52,9 @@ macro_rules! impl_for_fn {
         unsafe impl<$($ArgTy,)* R> ZeroableInOption for unsafe extern "C" fn($($ArgTy,)*) -> R {}
         unsafe impl<$($ArgTy,)* R> ZeroableInOption for extern "system" fn($($ArgTy,)*) -> R {}
         unsafe impl<$($ArgTy,)* R> ZeroableInOption for unsafe extern "system" fn($($ArgTy,)*) -> R {}
-        #[cfg(target_os="windows")]
+        #[cfg(all(target_os="windows", target_arch = "x86"))]
         unsafe impl<$($ArgTy,)* R> ZeroableInOption for extern "stdcall" fn($($ArgTy,)*) -> R {}
-        #[cfg(target_os="windows")]
+        #[cfg(all(target_os="windows", target_arch = "x86"))]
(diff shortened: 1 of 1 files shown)


Lokathor (approved):

Lokathor:
We'll call this a minor change, because theoretically breakage could occur, which makes it seem "bigger" than just a patch.

Lokathor:
released in 1.25

PR #341, opened 2025-12-09
Make derive(NoUninit) have the same constraints as Pod

### LINKED ISSUE #340: derive(NoUninit) on #[repr(transparent)] type with generic parameter?
I would like to have:
```
#[derive(NoUninit)]
#[repr(transparent)]
struct Wrapper<T>(T);
```
which seems like it should be sound if `T: NoUninit`. However it currently fails with:

```
error: NoUninit cannot be derived for structs containing generic parameters because the padding requirements can't be verified for generic structs
```

This does work for `Pod`.

Playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=518adfa01c2c70d7a7a31e28bc356247

jsgf:
Derive is allowed for generic types if they are `repr(transparent)` or `packed(1)`.

Also fixes the `completly_packed` typo because I didn't want to copy it and I didn't want the cases to be inconsistent.

Fixes #340

### DIFF
--- a/derive/src/traits.rs
+++ b/derive/src/traits.rs
@@ -66,20 +66,20 @@ impl Derivable for Pod {
   ) -> Result<TokenStream> {
     let repr = get_repr(&input.attrs)?;
 
-    let completly_packed =
+    let packed_or_transparent =
       repr.packed == Some(1) || repr.repr == Repr::Transparent;
 
-    if !completly_packed && !input.generics.params.is_empty() {
+    if !packed_or_transparent && !input.generics.params.is_empty() {
       bail!("\
-        Pod requires cannot be derived for non-packed types containing \
+        Pod cannot be derived for non-transparent/non-packed types containing \
         generic parameters because the padding requirements can't be verified \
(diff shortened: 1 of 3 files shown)


derive/src/traits.rs
```diff
@@ -257,14 +258,26 @@ impl Derivable for NoUninit {
   fn asserts(
     input: &DeriveInput, crate_name: &TokenStream,
   ) -> Result<TokenStream> {
-    if !input.generics.params.is_empty() {
-      bail!("NoUninit cannot be derived for structs containing generic parameters because the padding requirements can't be verified for generic structs");
+    let repr = get_repr(&input.attrs)?;
+
+    let completely_packed =
```
fu5ha: `packed_or_transparent` here as well

fu5ha (changes requested):
Just a few nits, otherwise I think this is a good change. Thanks :)

Lokathor:
@fu5ha want to do a final check?

fu5ha (approved):

PR #352, opened 2026-08-05
Bump bytemuck_derive to syn v3 and require rustc v1.71 for feature "derive"

Hommsin:
### Changes
- Upgrade the derive/proc-macro crate to syn v3 and set derive's declared MSRV to 1.71.

This change only affects the optional derive feature. The default build of bytemuck is unchanged.

### Motivation
syn v3 provides better support for newer Rust syntax and parsing bugfixes and reduces the need for multiple versions of this dependency.

### Tests performed
derive tests on 1.71: cargo +1.71.0 test --manifest-path=derive/Cargo.toml --all-targets --all-features
All tests passed.
Main crate test on current toolchain: cargo test --all-targets
All tests passed.

### Notes
Perhaps a version bump for the derive crate would also be in order.

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -9,7 +9,7 @@ keywords = ["transmute", "bytes", "casting"]
 categories = ["encoding", "no-std"]
 edition = "2018"
 license = "Zlib OR Apache-2.0 OR MIT"
-rust-version = "1.68"
+rust-version = "1.71"
 #note(lokathor): do not publish with this set
 #resolver = "3"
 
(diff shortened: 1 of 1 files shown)


Lokathor:
So this only needed to change the Cargo.toml? That's a lot simpler than I was expecting.

Hommsin:
Yeah I was surprised as well. It feels a little _too_ easy. Of course having the tests pass is not conclusive proof that there aren't edge cases that break under syn v3. But the derive crate's test look thorough to me.

Lokathor:
released 1.12 of the derives

PR #354, opened 2026-09-09
Forward struct where-clauses on TransparentWrapper derive

teddytennant:
`#[derive(TransparentWrapper)]` drops the struct's own where-clause because TransparentWrapper sets `requires_where_clause()` to false and the derive then emits `None` instead of the original predicates. A `struct A<T>(T) where T: Trait` therefore fails E0277 on the generated `unsafe impl`.

Snapshot the clause before `add_trait_marker` and forward it on that path. Compile-pass coverage is `TransparentWithWhereClause` in `derive/tests/basic.rs`.

`cargo test --manifest-path derive/Cargo.toml` and `cargo test --lib --tests` are green. Unfixed main still fails the E0277 repro.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,6 +12,7 @@ jobs:
     name: Test Rust ${{ matrix.rust }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         # it's a little tempting to use `matrix` to do a cartesian product with
         # our `--feature` config here, but doing so will be very slow, as the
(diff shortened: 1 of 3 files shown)


teddytennant:
The matrix fail-fasted on `stable-i686-gnu` (windows): rustup 1.29.1 now refuses that host toolchain (`may not be able to run on this system`) before cargo runs. Same job was green on main in August (7cc6738) and on #353. The 1.71 / windows-stable logs that did start still show `test result: ok`.

Lokathor:
There's a successful i686-unknown-linux-gnu, but we may have to just adjust CI.

I do recall them announcing that i686 is going down to tier2,  so it can probably be set up as a cross-check of some sort.

I won't be at my laptop and able to have a more serious look until at least several hours from now. If you want to try and adjust the CI settings you can make that part of this PR.

teddytennant:
dropped the windows i686 host toolchain cells (rustup 1.29 refuses them on 64-bit runners) and set fail-fast: false. 32-bit stays on the linux cross job.

Lokathor:
`bytemuck_derive-v1.12.1` released

PR #355, opened 2026-09-14
Add trivial case to TransparentWrapper

Cyclic3:
I am by no means an expert on rust safety, but it seems to me that every type should be able to implement `TransparentWrapper<Self>`. This may seem kinda dumb, but I've been working on some units-of-measure code where being able to talk about "things that can be an `impl ComplexFloat`" without having to worry whether I've just got an unwrapped type would be *very* helpful in cleaning things up.

Please feel free to tell me I'm stupid if I've misunderstood the safety assumptions in implementing this!


Cyclic3:
As a heads up, this PR would mean that `Type::peel` is now ambiguous whenever you implement TransparentWrapper, which will definitely ruin someone's day. I've marked it as a draft accordingly.

Lokathor:
I don't think we want to ruin anyone's day.

could you show an example of code that would break?

and can you show an example of your code and how it gets cleaned up by this impl?

maybe we can add this as a cargo feature opt-in, if the breakage isn't too big and the cleanup is significant.

Cyclic3:
My actual code is rather messy and macro heavy, so I'll try and make a simplified version here; I'm really sorry for the length!

I want to mark various number types as having units:
```rs
/// Magic trait that tracks unit conversions
trait Unit { /* ... */}

/// Marker trait for something that has units
trait United {
    type Inner;
    type Unit: Unit;
}

#[derive(Debug, bytemuck::TransparentWrapper, PartialEq, Eq, PartialOrd, Ord)]
#[transparent(T)]
#[repr(transparent)]
struct UnitWrapper<T, U: Unit>(T, PhantomData<U>);
impl<T, U: Unit> United for UnitWrapper<T, U> {
    type Inner = T;
    type Unit = U;
}
impl<T: num_complex::ComplexFloat> United for T {
    type Inner = Self;

    type Unit = base_units::Dimensionless;
}
```
I then want to be able to, for example, multiply quantities with units in a way that respects those units:
```rs
impl<LhsT, LhsU: Unit, Rhs: United + bytemuck::TransparentWrapper<Rhs::Inner>> std::ops::Mul<Rhs> for UnitWrapper<LhsT, LhsU>
    where LhsT: std::ops::Mul<Rhs::Inner>, LhsU: std::ops::Mul<Rhs::Unit>, LhsU::Output: Unit
{
    type Output = UnitWrapper<LhsT::Output, LhsU::Output>;

    fn mul(self, rhs: Rhs) -> Self::Output {
        Self::Output::wrap(self.0.mul(Rhs::peel(rhs)))
    }
}
```
This all works fine if we stay within our wrapper class:
```rs
pub type Time<T> = UnitWrapper<T, unit!(Samples)>;
pub type Freq<T> = UnitWrapper<T, unit!(Samples^-1)>;
#[test]
fn time_freq_cancellation() {
    let x = Time::wrap(24f32);
    let y = Freq::wrap(1.5f32);
    let z/*: UnitWrapper<f32, base_unit::Dimensionless>*/ = x * y;
    assert_eq!(UnitWrapper::peel(z), 32.);
}
```
But if I want to multiply by scalars, this goes bad:
```rs
#[test]
fn time_scaling() {
    let x = Time::wrap(6.25f32);
    let z = x * 10.72; // will fail, as f32 does not implement TransparentWrapper
    let z_value = Time::peel(z);
    assert_eq!(z_value, 67.)
}
```
It would appear that I can solve this problem for creating a duplicate implementation for every `ComplexFloat`, but rust doesn't like that because I could implement all of `United`, `ComplexFloat` and `TransparentWrapper` on the same type in the future.

As far as I can see it, I am therefore left with the following options:
1. Require the user to write something like `Dimensionless(10.72)` each time they want to use a scalar (unergonomic and makes it a pain to use).
2. Re-expose the methods of `TransparentWrapper` in `United` (brittle and ugly).
3. Implement this for every concrete `ComplexFloat` type (horrifyingly brittle and locked)
4. Make this PR :)

It's worth noting out that option 3 doesn't make this code work fully either. In fact, both tests break because `UnitWrapper::peel` is ambiguous between `TransparentWrapper<Self>` and `TransparentWrapper<f32>`, and so I either have to explicitly constrain the output, or do one of the ugly `<UnitWrapper as TransparentWrapper<f32>>::peel` casts. Hence why I've left this as a draft.

Lokathor:
Couldn't you just write an impl like
```rust
impl Mul<f32> for Time<f32> { ... }
```
So that you can scale values by the unwrapped type that way?

Cyclic3:
Yeah, that's my option 3: I'd have to do that for every concrete type (I want to allow different float precisions and have a couple integers). It's not that my code doesn't work without this PR, it's just brittle and prevents users using their own types (i.e. something like a bigint) without having to faff with NewTypes.

